PR #2, opened 2015-10-27
adding more functionality

loganwright:


src/Async/Qu.swift
```diff
@@ -0,0 +1,200 @@
+//
+//  Qu.swift
+//
+//  Created by Logan Wright on 10/24/15.
+//  Copyright © 2015 lowriDevs. All rights reserved.
+//
+
+public typealias Block = () -> ()
```
brightredchilli, line 8: is it not dangerous to do a public typealias without prefix here?

src/Async/Qu.swift
```diff
@@ -0,0 +1,200 @@
...
+}
+func +=(operationQueue: NSOperationQueue, operation: NSOperation) {
+    operationQueue.addOperation(operation)
+}
+
+// MARK: NSOperationQueue+Operations
+
+private extension NSOperationQueue {
+    
+    /// This is the most recently added operation.  According to the docs, `operation` is returned in the order they were added to the queue, NOT the order in which they are executed.
```
brightredchilli, line 179: you mean 'operations' here?

src/Async/Qu.swift
```diff
@@ -0,0 +1,200 @@
...
+    }
+    
+    func Then(block: Block) -> Self {
+        return ThenAfter(.Last, block: block)
+    }
+    
+    func ThenAfter(dependency: Dependency, block: Block) -> Self {
+        let blockOp = Operation(block: block)
+        switch dependency {
+        case .Last:
```
brightredchilli, line 94: might be a formatting foobar here, ps...

src/Colors/ColorDescriptor.swift
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+    
+    // MARK: Initializers
+    
+    init(unicodeScalarLiteral value: UnicodeScalarLiteralType) {
+        self.init(value)
+    }
+    
+    init(extendedGraphemeClusterLiteral value: ExtendedGraphemeClusterLiteralType) {
```
brightredchilli, line 55: wow these are painful

src/StandardLibrary/String/String+Substrings.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+
+extension String {
+    subscript(range: Range<Int>) -> String {
+        let chars = Array(characters)
+        let substringCharacters = chars[range]
+        return String(substringCharacters)
+    }
+    
+    mutating func dropFirst() {
```
brightredchilli, line 18: guard if out of bounds, throws? :monkey:

src/Async/Qu.swift
loganwright, line 94: This is how switch statements are auto-formatted in Swift, I generally don't fight it :)

src/Colors/ColorDescriptor.swift
loganwright, line 55: Yea, it's nice once they're setting them up, but getting it going is a pain

src/StandardLibrary/String/String+Substrings.swift
loganwright, line 18: Good call, I don't think throwing is as appropriate as a guard, I'll add that in!

src/Async/Qu.swift
loganwright, line 179: Yup, fixed.

src/Async/Qu.swift
loganwright, line 8: Nope, typealiases are also namespaced, so if there's a conflicting `Block`, this one will be `IntrepidSwiftWisdom.Block`

brightredchilli, line 8: Cool. So if we start including a new library that is also called block, all our existing references for that project will have to be renamed, something like that?

loganwright, line 8: If there's two libraries competing for the typealias, yes we'd have to replace them with a namespace if used.

The main reason I have it as public is that I'm using this to depict public argument types, so it kind of needs to be publicly exposed. Do you think I should just remove it and write all of the arguments out?

brightredchilli, line 8: nope. i think at this point we dont really know what the real problems are. you defiitely have a better understanding of this than me. I think we should take this as a note to ourselves, and at the first point where we see there is a problem we should get this modified.

PR #5, opened 2015-10-28
Logan - More components

loganwright:
Not sure if the view extension stuff is the best implementation, but I find myself rewriting `CGRectGetWidth` a million times per project, and I think it'll be helpful.


src/Async/After/After.swift
```diff
@@ -0,0 +1,34 @@
...
+
+public func After(after: NSTimeInterval, op: () -> (), completion: (() -> Void)?) {
+    let seconds = Int64(after * Double(NSEC_PER_SEC))
+    let dispatchTime = dispatch_time(DISPATCH_TIME_NOW, seconds)
+    dispatch_after(dispatchTime, dispatch_get_main_queue()) {
+        let blockOp = NSBlockOperation(block: op)
+        blockOp.completionBlock = completion
+        NSOperationQueue.mainQueue().addOperation(blockOp)
+    }
+}
```
paulrolfe: new line at end of file if you wanna, not that important.

src/Geometry/Geometry.swift
```diff
@@ -0,0 +1,91 @@
...
+    public var radians: CGFloat {
+        return CGFloat(Double(self) * (M_PI / 180))
+    }
+}
+
+public extension CGFloat {
+    public var degrees: CGFloat {
+        return self * CGFloat(180.0 / M_PI)
+    }
+}
```
paulrolfe, line 21: Curious why you don't combine this with the `CGFloat` extension above it? I see this a lot.

loganwright, line 21: Because I made them at different times, and threw them all in this file :)  Will fix.

paulrolfe:
:goat: :seedling:

src/Async/After/After.swift
```diff
@@ -0,0 +1,34 @@
...
+        wrappedCompletion = completion
+    }
+    
+    After(after, op: op, completion: wrappedCompletion)
+}
+
+public func After(after: NSTimeInterval, op: () -> (), completion: (() -> Void)?) {
+    let seconds = Int64(after * Double(NSEC_PER_SEC))
+    let dispatchTime = dispatch_time(DISPATCH_TIME_NOW, seconds)
+    dispatch_after(dispatchTime, dispatch_get_main_queue()) {
```
brightredchilli: I feel we should guard that we are on the main thread when this method gets called. or we can rename the block to completion on main thread. this behaviour is sort of undocumented in this library.

src/UIKit/UIView+Geometry.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  UIView+Geometry.swift
+//
+//  Created by Logan Wright on 10/24/15.
+//  Copyright © 2015 lowriDevs. All rights reserved.
+//
+
+public extension UIView {
```
brightredchilli: i feel like these methods play a lot nicer as cgrect extensions than uiview extensions.

then 
view.frame.width
view.bounds.halfwidth

just works without us having to define separate methods. and i think it adds some readability.

src/Async/After/After.swift
loganwright: Good point, do you think it should run on the calling thread, or document main, or provide an option to specify and maybe default to main queue

src/UIKit/UIView+Geometry.swift
loganwright: That was exactly my concern!  I'll swap them out

PR #6, opened 2015-11-09
Add Optional Handling to XCAssert Methods

cprime:
Suppose you have two optionals, `value1` and `value2`, and you want to test their equality:

Calling `XCTAssertEqual(value1, value2)` is unsupported because `XCTAssertEqual` does not take optionals.

Calling `XCTAssertEqual(value1!, value2!)` is unsupported because unwrapping `value1` and `value2` will cause the test run stopping crash.

With these additions, calling `ip_XCTAssertEqualOptional(value1, value2)` will first check whether or not the optionals have a value, and it they do, it asserts the equality of their unwrapped values.

### DIFF
--- a/src/XCTest/XCTest+Optionals.swift
+++ b/src/XCTest/XCTest+Optionals.swift
@@ -1,0 +1,30 @@
+import Foundation
+import XCTest
+
+public func ip_XCTAssertEqualOptional<T: Any where T: Equatable>(@autoclosure expression1: () -> T?, @autoclosure expression2: () -> T?, _ message: String? = nil, file: String = __FILE__, line: UInt = __LINE__) {
+    let exp1 = expression1()
+    let exp2 = expression2()
+    switch (exp1, exp2) {
+    case let (left?, right?):
+        XCTAssertEqual(left, right, message ?? "", file: file, line: line)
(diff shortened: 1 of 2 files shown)


src/XCTest/XCTest+Optionals.swift
```diff
@@ -0,0 +1,26 @@
...
+            XCTAssertEqual(exp1, exp2, (message != nil ? message! : ""), file: file, line: line)
+        } else {
+            XCTFail((message != nil ? message! : "exp1 != nil, exp2 == nil"), file: file, line: line)
+        }
+    } else if let _ = expression2() {
+        XCTFail((message != nil ? message! : "exp1 == nil, exp2 != nil"), file: file, line: line)
+    }
+}
+
+public func ip_XCTAssertEqualWithAccuracy<T : FloatingPointType>(@autoclosure expression1: () -> T?, @autoclosure _ expression2: () -> T?, accuracy: T, _ message: String? = nil, file: String = __FILE__, line: UInt = __LINE__) {
```
loganwright: Should this one have the `Optional` specifier in function name like above?

loganwright:
:shipit:

Issue #4, opened 2015-10-27
Setup cataloging of components


loganwright:
- [x] Document what's available 
- [x] add basic usage of more complex components.

PR #8, opened 2015-11-10
more components and documentation

loganwright:
#4


README.md
```diff
@@ -8,7 +8,136 @@ pod 'IntrepidSwiftWisdom'
...
+    var color: UIColor {
+        return rawValue.color
+    }
+}
+```
+
+And use:
+
+```Swift
+someView.backgroundColor = ColorPalette.White.color
```
brightredchilli, line 106: I feel like its worth noting somewhere that zeplin does generate swift color palletes, though this might not be the right place for it

README.md
```diff
@@ -8,7 +8,136 @@ pod 'IntrepidSwiftWisdom'
...
+enum Setting : String, EnumSettingsKeyAccessible {
+    case DisplayName
+    case LastOpenDate
+}
+```
+
+##### Read
+
+```Swift
+let displayName: String? = Setting.DisplayName.readFromDefaults
```
brightredchilli, line 144: hmm, kind of want this to be Setting.DisplayName with a getter and setter.

This is cool though, let's see how it gets adopted and used.

src/Foundation/Defaults.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import Foundation
+
+protocol SettingsKeyAccessible {
+    var key: String { get }
+    var defaults: NSUserDefaults { get }
+}
+
+extension SettingsKeyAccessible {
+    var defaults: NSUserDefaults {
```
brightredchilli: I would make this a convenience initializer with nsuserdefaults to be passed in as the default so we pave the way for testing

src/UIKit/CellConfiguring.swift
```diff
@@ -0,0 +1,71 @@
...
+    
+    func dequeueHeader<T: UITableViewHeaderFooterView>(section: Int, identifier: String = T.identifier) -> T {
+        let header = dequeueReusableHeaderFooterViewWithIdentifier(T.identifier) as! T
+        return header
+    }
+}
+
+extension UITableViewHeaderFooterView {
+    class var nibName: String {
+        let name = "\(self)".componentsSeparatedByString(".").first ?? ""
```
brightredchilli: what is this for? seems comment worthy

src/UIKit/CellConfiguring.swift
```diff
@@ -0,0 +1,71 @@
...
+        return name
+    }
+    class var nib: UINib? {
+        if let _ = NSBundle.mainBundle().pathForResource(nibName, ofType: "nib") {
+            return UINib(nibName: nibName, bundle: nil)
+        } else {
+            return nil
+        }
+    }
+    class var identifier: String {
```
brightredchilli: always nervous about these things - should we prefix to avoid potential collisions?

src/UIKit/CellConfiguring.swift
```diff
@@ -0,0 +1,71 @@
...
+        return "identifier:\(self)"
+    }
+}
+
+extension UITableViewCell {
+    class var nibName: String {
+        let name = "\(self)".componentsSeparatedByString(".").first ?? ""
+        return name
+    }
+    class var nib: UINib? {
```
brightredchilli: oh you're not gonna make uitableviewcell and uitableviewheaderfooterview extend the NibNameGettableAndIdentifiable protocol??

just kidding, just kidding

README.md
loganwright, line 144: Maybe I'll edit the example, but that's how I'm actually using it.  In the project this was made for, the enums are private and I have something like:

```
struct ApplicationSettings {
    var displayName: String {
        get {
            return Setting.DisplayName.readFromDefaults ?? ""
        }
        set {
            Setting.DisplayName.writeToDefaults(newValue)
        }
    }
}
```

Do you think it's worth mentioning that in readme?

src/Foundation/Defaults.swift
loganwright: I think you could test this by having either a mock protocol, or you could define an implementation for each target that overrides this in a different way.

Adding injection would add more requirements for implementation I think.

src/UIKit/CellConfiguring.swift
loganwright: Ahh, yup, I'll do that!

src/UIKit/CellConfiguring.swift
loganwright: Don't tempt me!

src/UIKit/CellConfiguring.swift
loganwright: Yea, definitely.  I'll update all of the files I just added.

README.md
loganwright, line 106: Noted in the readme.

PR #10, opened 2015-11-10
Create separate subspec for testing targets

andrewdolce:
This fixes the issue caused by linking XCTest into the main target. Logan will be updating the README to explain how to include the testing code into testing targets, but in a nutshell:

You should include the default subspec for your main target: `pod IntrepidSwiftWisdom`.
You should include the testing subspec for your test target: `pod IntrepidSwiftWisdom/Testing`.


loganwright:
Since we haven't yet pushed the `0.0.7` tag, to keep things ordered, do you want to make this `0.0.6`?

Otherwise, :ship:

andrewdolce:
Sure, I'll make it 0.0.6. But I'm actually not convinced this is completely working yet. We need to make sure that we can actually use the `/Testing` subspec in a testing target.

loganwright:
I think we'll want to go w/ #11 based on test target integration issues related to cocoapods.  I'd like to revisit subspecs in the future.

PR #12, opened 2015-11-13
adding result type

loganwright:
result type and time of day


SwiftWisdom/SwiftWisdom/Core/Foundation/Time/TimeOfDay.swift
```diff
@@ -0,0 +1,100 @@
...
+    
+    public var displayString: String {
+        // Backing Format
+        dateFormatter.dateFormat = "HH:mm"
+        guard let date = dateFormatter.dateFromString(stringRepresentation) else {
+            return ""
+        }
+
+        // Display format
+        dateFormatter.dateFormat = "h:mm a"
```
brightredchilli: shouldnt this be displayFormat?

SwiftWisdom/SwiftWisdom/Core/Foundation/Time/TimeOfDay.swift
```diff
@@ -0,0 +1,100 @@
...
+ *  A struct represention of dates expressed as a time of day.
+ */
+public struct TimeOfDay {
+    
+    // MARK: Public Properties
+    
+    public let hours: Int
+    public let minutes: Int
+    
+    public var displayFormat: String = "h:mm a" // 9:18 am
```
brightredchilli, line 21: setting displayFormat might need to be reflected down into the private date formatter

SwiftWisdom/SwiftWisdom/Core/Foundation/Time/TimeOfDay.swift
loganwright: Yup, good catch.

PR #13, opened 2015-12-10
adding percent animator for layer animations that don't work with tra…

loganwright:
…dational ca or uiview animations

### DIFF
--- a/SwiftWisdom/SwiftWisdom/Core/Downloader/Downloader.swift
+++ b/SwiftWisdom/SwiftWisdom/Core/Downloader/Downloader.swift
@@ -1,0 +1,84 @@
+//
+//  Downloader.swift
+//  bmap
+//
+//  Created by Logan Wright on 12/2/15.
+//  Copyright © 2015 Intrepid. All rights reserved.
+//
+
+import UIKit
+
+public final class Downloader : NSObject, NSURLSessionDownloadDelegate {
+    
+    public enum State {
+        case Began
+        case InProgress(CGFloat)
+        case Completed(Result<NSURL>)
+    }
+    
+    public let url: NSURL
+    public let id = NSUUID().UUIDString
+    
+    private var updater: (State -> Void)?
+    
+    private lazy var session: NSURLSession =
+    NSURLSession(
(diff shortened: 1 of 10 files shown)


SwiftWisdom/SwiftWisdom/Core/Downloader/Downloader.swift
```diff
@@ -0,0 +1,84 @@
...
+        updater?(.InProgress(downloadTask.percentageDownloaded))
+    }
+    
+    public func URLSession(session: NSURLSession, downloadTask: NSURLSessionDownloadTask, didResumeAtOffset fileOffset: Int64, expectedTotalBytes: Int64) {
+        updater?(.Began)
+        updater?(.InProgress(downloadTask.percentageDownloaded))
+    }
+}
+
+func ==(lhs: Downloader, rhs: Downloader) -> Bool {
```
markdanyo, line 69: Should the `Equatable` extension declaration be provided here?

loganwright, line 69: Since this inherits from `NSObject`, it technically already conforms to `Equatable`, so I get warnings on redeclaration.  This just overrides the implementation to use mine instead of the NSObject base.

SwiftWisdom/SwiftWisdom/Core/Version/Version.swift
```diff
@@ -0,0 +1,81 @@
...
+import Foundation
+
+public struct Version {
+    public private(set) var major: Int = 0
+    public private(set) var minor: Int = 0
+    public private(set) var patch: Int = 0
+    
+    init?(string: String) {
+        let components = string
+            .componentsSeparatedByString(".")
```
paulrolfe, line 18: Is this new line a part of the style guide?

loganwright, line 18: Not explicitly, but we should look for a nice way to allow for newline situations.  This one is possibly a bit of a stretch, but often it's a good way to keep lines compact.

PR #15, opened 2015-12-15
Pr/points sizes

paulrolfe:
more math for points and sizes


loganwright:
Do you want to add a tag and update the Podspec?  Then :shipit:

PR #16, opened 2015-12-26
Previous GetConstraints had a lot of errors.

paulrolfe:
I verified it works this time in bose-connect.


loganwright:
I think this might be a good opportunity to put some quick tests in around the issue if they wouldn't be too costly.

I didn't take another pass at the existing code, but maybe some of the comments can be applied elsewhere too.

Update Podspec, Tag, and :shipit:

SwiftWisdom/SwiftWisdom/Core/UIKit/UIView+GetConstraints.swift
```diff
@@ -59,27 +59,32 @@ public extension UIView {
...
+     :param: attribute     Attribute to look for belonging to the receiver
+     :param: view1        Defaults to the receiver, but could be any subview as well.
+     :param: view2        Probably a UIView, that view1 is related to.
+     :param: viewAttribute Attribute that should be used with the toItem
+     
+     :returns: The first constraint that matches. May return unexpected constraint if receiver contains multiple constraints with this item and itemAttribute.
+     */
+    public func ip_constraintForAttribute(attribute: NSLayoutAttribute, var onView view1: UIView? = nil, toView view2: UIView? = nil, viewAttribute: NSLayoutAttribute? = nil) -> NSLayoutConstraint? {
+        
+        if view1 == nil {
```
loganwright: I think instead of this, and the subsequent `!`, I'd rather see an unwrapped `let` here.  The `var` is actually going to be removed in the next Swift release from arguments, so better to change this now.  After reading the method, I also think the external parameters make for better internal names since they're more clear.

Make internal argument `onView`  as well.  You can unwrap to same name which is common for these situations

``` Swift
let onView = onView ?? self
```

Also make `view2`, `toView` internally

SwiftWisdom/SwiftWisdom/Core/UIKit/UIView+GetConstraints.swift
```diff
@@ -59,27 +59,32 @@ public extension UIView {
...
+            view1 = self
+        }
         
-        if let toAttribute = itemAttribute, let toItem = view2 {
+        if let toAttribute = viewAttribute, let toItem = view2 {
             
             return constraints.filter { constraint in
-                return constraint.ip_relatesView(view: view1, viaAttribute: attribute, toView: toItem, andItsAttribute: toAttribute)
+                return constraint.ip_relatesView(view: view1!, viaAttribute: attribute, toView: toItem, andItsAttribute: toAttribute)
```
loganwright: Avoid unwrapping in this way in general, save `!` unwrapping only for a last resort.  Even though the current implementation, we can say that it will definitely exist, we shouldn't rely on it by human code comprehension.  We should let Swift handle it for us, so it can't reach this line of code unless `onView` (`view1`) has value.  This allows us to rely on its implementation here regardless of what happens above in future implementations since it won't build unless that condition is satisfied.

SwiftWisdom/SwiftWisdom/Core/UIKit/UIView+GetConstraints.swift
```diff
@@ -59,27 +59,32 @@ public extension UIView {
...
                 }
                 .first
             
         } else if let toItem = view2 {
             
             return constraints.filter { constraint in
-                return constraint.ip_relatesView(view: view1, viaAttribute: attribute, toView: toItem)
+                return constraint.ip_relatesView(view: view1!, viaAttribute: attribute, toView: toItem)
```
loganwright: Same here n below

PR #17, opened 2016-01-07
Pr/timeout

paulrolfe:
make it public


loganwright:
Nice, thanks Paul!

:shipit: 

Did we ever set you up to push to intrepids spec host?

paulrolfe:
No we never did that

PR #18, opened 2016-01-20
Lw/additions

loganwright:
I'm adding a bunch of stuff here from some of the work we've done on Bose. Lots of NSData material in particular.

I also changed the folder structure and removed the weird pods integration things since it wasn't really useful and added complexity.  

Still needs:

Possibly breaks out NSData stuff into a subspec.
Add more readme explanations at some point


SwiftWisdom/Core/StandardLibrary/Numbers/Bool+Extensions.swift
```diff
@@ -0,0 +1,22 @@
...
+extension Bool {
+    public init(ip_string: String) {
+        let cleaned = ip_string.lowercaseString
+        let truePrefixes = [
+            "t",
+            "y",
+            "1",
+        ]
+        
+        self = truePrefixes.contains { cleaned.hasPrefix($0) }
```
paulrolfe: Is it widely accepted that strings beginning with y, t, or 1 represent a true value? I've never heard that.

SwiftWisdom/Core/UIKit/UIView+Geometry.swift
```diff
@@ -0,0 +1,229 @@
...
+//
+//  Created by Logan Wright on 1/19/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import UIKit
+
+extension UIView {
+    public var ip_width: CGFloat {
+        return CGRectGetWidth(frame)
```
paulrolfe: not bounds?

SwiftWisdom/Core/UIKit/UIView+Geometry.swift
```diff
@@ -0,0 +1,229 @@
...
+    }
+    public var ip_halfWidth: CGFloat {
+        return ip_width / 2.0
+    }
+    public var ip_height: CGFloat {
+        return CGRectGetHeight(frame)
```
paulrolfe: not bounds?

SwiftWisdom/Core/StandardLibrary/Numbers/Bool+Extensions.swift
loganwright: This is how it was done in `Foundation`:

screen shot 2016-01-20 at 12 01 10 am

I think as long as a user is expecting to get a bool from a string, it should function as expected.

SwiftWisdom/Core/UIKit/UIView+Geometry.swift
loganwright: for width, bounds and height should always be equal.  For whatever reason though, bounds feels more right here, so I'll swap it out.

brightredchilli: bounds is right here, because ip_frameWidth is defined below. However, I'm finding these getters to be a little dated since frame has default getters for say, width and height now

SwiftWisdom/Core/UIKit/UIViewController+Nibs.swift
```diff
@@ -0,0 +1,22 @@
...
+import UIKit
+
+extension UIViewController {
+    public static func ip_fromNib() -> Self {
+        let controller = self.init(nibName: ip_nibName, bundle: nil)
+        return controller
+    }
+}
+
+extension UIViewController {
```
brightredchilli: any reason these are not in the same extension?

SwiftWisdomTests/NSDataSegmentGeneratorTests.swift
```diff
@@ -0,0 +1,72 @@
...
+class NSDataSegmentGeneratorTests: XCTestCase {
+
+    let dataPath = NSBundle(forClass: NSDataSegmentGeneratorTests.self)
+        .pathForResource("mock_firmware_update", ofType: ".dfu")!
+    
+    lazy var originalData: NSData = NSData(contentsOfFile: self.dataPath)!
+    
+    let segmentLength = 100
+    
+    override func setUp() {
```
brightredchilli: unused methods, maybe remove

SwiftWisdomTests/NSDataSegmentGeneratorTests.swift
```diff
@@ -0,0 +1,72 @@
...
+            if let compareSuffix = compareSuffix {
+                XCTAssert(reassembly == compareSuffix)
+            } else {
+                XCTAssert(reassembly.length == 0)
+                XCTAssert(compareSuffix == nil)
+            }
+        }
+    }
+    
+    func testPerformanceExample() {
```
brightredchilli: unused remove

SwiftWisdom/Core/UIKit/UIViewController+Nibs.swift
loganwright: Should be, moving now.  They came from different times in a project and I hadn't combined them until now.

SwiftWisdom/Core/UIKit/UIView+Geometry.swift
loganwright: @brightredchilli I'm really torn on this one, there's a couple considerations. My main one is that whenever I'm doing any programmatic layout calculations, I use these methods so much that even with the other accessors, I always end up adding them.

I think the main motivation for the `frameX`, `boundsMidX` naming stuff is because setters are involved, so we need to be a bit more articulate.

With the properties here, they're access only so I think they're safe enough to include.

Thoughts on this?

brightredchilli: I'm perfectly fine leaving them in. I do think however, that once you have ip_width and ip_frameWidth, it gets quite close to view.bounds.width and view.frame.width. I think we can always add them, since one can always choose not to use them!

SwiftWisdom/Core/Foundation/Not.swift
```diff
@@ -0,0 +1,18 @@
...
+//  Created by Logan Wright on 12/11/15.
+//  Copyright © 2015 Intrepid. All rights reserved.
+//
+
+/**
+ Inverts a given boolean return closure
+ 
+ Good for things like `array.filter(not(matchesThing))`
+ */
+public func not<T>(original: T -> Bool) -> T -> Bool {
```
brightredchilli, line 14: quick question, if we happen to include a library that exposes not as a global function, is there a way we can 'back out'?

loganwright, line 14: Global functions are namespaced, so if two modules have declare `not`, we would have to specify like: `IntrepidSwiftWisdom.not()` or `SomeOtherModule.not(`

brightredchilli:
looks good :ship:

PR #19, opened 2016-01-26
fixing cell dequeueing

loganwright:


SwiftWisdom/Core/UIKit/CellConfiguring.swift
```diff
@@ -26,12 +26,12 @@ public extension UITableView {
     }
     
     public func ip_dequeueCell<T: UITableViewCell>(indexPath: NSIndexPath, identifier: String = T.ip_identifier) -> T {
-        let cell = dequeueReusableCellWithIdentifier(T.ip_identifier, forIndexPath: indexPath) as! T
```
brightredchilli, line 29: whats the reason for this change?

loganwright, line 29: The original implementation was ignoring the `identifier` argument regardless of what was specified, so if you registered your value with an identifier different from the auto-generated one, it would fail.

Ultimately, making sure that if we expose an `identifier` argument, we're actually using it

brightredchilli:
:ship:

PR #20, opened 2016-01-31
Some additions / Fixes / Testing

loganwright:


SwiftWisdom/Core/Colors/ColorDescriptor.swift
```diff
@@ -64,18 +79,25 @@ extension ColorDescriptor : StringLiteralConvertible, RawRepresentable, Equatabl
         self.init(rawValue)
     }
     
-    init(_ string: String) {
+    public init(_ string: String) {
         let rgbComponents = string.componentsSeparatedByString(",")
-            .flatMap { Int($0) }
         if rgbComponents.count == 4 {
-            self = .RGB(r: rgbComponents[0], g: rgbComponents[1], b: rgbComponents[2], a: rgbComponents[3])
+            if rgbComponents.first?.containsString(".") == true {
```
paulrolfe: What if it's just `0`? i.e. rgb values of `0, 0.8, 0.4, 1`

SwiftWisdom/Core/UIKit/UIImage+Overlay.swift
```diff
@@ -0,0 +1,29 @@
...
+    public final func ip_overlayImage(overlayImage: UIImage, inRect innerRect: CGRect) -> UIImage {
+        let outerRect = CGRect(origin: CGPointZero, size: size)
+        UIGraphicsBeginImageContext(outerRect.size)
+        self.drawInRect(outerRect)
+        overlayImage.drawInRect(innerRect)
+        let img = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+        return img
+    }
+}
```
paulrolfe: new line at end.

SwiftWisdom/Core/Colors/ColorDescriptor.swift
loganwright: The alternative I was toying with is checking the entire string for a `.`, that might be a better approach.

paulrolfe: Yeah, I think that's a good idea.

paulrolfe:
:fire: :arrow_up:

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -18,8 +18,9 @@ extension Int {
     }
 }
 
-extension Int {
+extension IntegerType {
     public func times(@noescape closure: Block) {
```
paulrolfe: Should this be `ip_times` to avoid ambiguity?

loganwright: yup, good call

PR #21, opened 2016-02-08
NSRange and Attributed Strings

paulrolfe:
Add some tests and fixed some tests.

### DIFF
--- a/SwiftWisdom/Core/StandardLibrary/NSAttribitedString/NSMutableAttributedString+Format.swift
+++ b/SwiftWisdom/Core/StandardLibrary/NSAttribitedString/NSMutableAttributedString+Format.swift
@@ -1,0 +1,63 @@
+//
+//  NSAttributedString+Format.swift
+//  bose-connect-ios
+//
+//  Created by Paul Rolfe on 1/16/16.
+//  Copyright © 2016 Intrepid-Pursuits. All rights reserved.
+//
+
+import Foundation
+
+extension NSMutableAttributedString {
+    
+    /**
+     Create a mutable attributed string with the attributes applied to all arguments.
+     
+     - parameter string:     formatted string where `%@` represents an argument.
+     - parameter attributes: attributes dictionary to use on arguments
+     - parameter args:       comma separated list of arguments
+     
+     - returns: NSMutableAttributedString with the attributes applied to argument strings.
+     */
+    convenience init(stringWithFormat string: String, applyingAttributes attributes: [String : AnyObject], toArgs args: String...) {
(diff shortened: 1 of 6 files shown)


SwiftWisdom/Core/StandardLibrary/NSAttribitedString/NSMutableAttributedString+Format.swift
```diff
@@ -0,0 +1,63 @@
...
+    
+    /**
+     Create a mutable attributed string with the given collection of NSAttributedString
+     
+     - parameter string: formatted string where `%@` represents an argument.
+     - parameter args:   comma separated list of arguments (NSAttributedString)
+     
+     - returns: NSMutableAttributedString with arguments inserted according to format.
+     */
+    private convenience init(stringWithFormat string: String, withArguments args: [NSAttributedString]) {
```
loganwright: any reason to have this `private`? Might be useful to be able to call w/ an array as well.  I might change the above `_ args` to match this signature so the overload is clear that they can use either.  

Overloading w/ array seems to be convention until splatting is explicitly supported

SwiftWisdomTests/SwiftWisdomTests.swift
```diff
@@ -28,3 +28,30 @@ class GetConstraintTests: XCTestCase {
...
+            attributes: [NSForegroundColorAttributeName : UIColor.redColor()]
+        )
+        let testString = NSMutableAttributedString(stringWithFormat: "This color is %@", redText)
+        let otherTestString = NSMutableAttributedString(
+            stringWithFormat: "This color is %@",
+            applyingAttributes: [NSForegroundColorAttributeName : UIColor.redColor()],
+            toArgs: "RED"
+        )
+        
+        XCTAssertEqual(testString, otherTestString)
```
loganwright, line 50: Does this test the substring colors? ie: if we made one `blueColor`? would it fail?

loganwright:
Cool, good additions.

Add `ip_` prefix to all functions and :shipit:

SwiftWisdomTests/SwiftWisdomTests.swift
paulrolfe, line 50: Yup, but I like that test, I added it in.

SwiftWisdom/Core/StandardLibrary/NSAttribitedString/NSMutableAttributedString+Format.swift
paulrolfe: I just didn't know overloading was preferred. I'll do that.

PR #22, opened 2016-02-16
Mathable

paulrolfe:
- Mean, median, mode of mathable collections -- they return their element type.
- Sequence extensions for most common elements, unique elements, and number of each element.

### DIFF
--- a/SwiftWisdom/Core/Geometry/Geometry.swift
+++ b/SwiftWisdom/Core/Geometry/Geometry.swift
@@ -86,6 +86,11 @@ public func random(range: Range<Int>) -> Int {
 }
 
 public func avg(numbers: CGFloat...) -> CGFloat {
+    return avg(numbers)
+}
+
+public func avg(numbers: [CGFloat]) -> CGFloat {
+    guard numbers.count > 0 else { return 0 }
     return numbers.reduce(0, combine: +) / CGFloat(numbers.count)
 }
 
--- a/SwiftWisdom/Core/Geometry/Geometry.swift
+++ b/SwiftWisdom/Core/Geometry/Geometry.swift
@@ -85,15 +85,6 @@ public func random(range: Range<Int>) -> Int {
     return random
 }
 
-public func avg(numbers: CGFloat...) -> CGFloat {
-    return avg(numbers)
-}
-
-public func avg(numbers: [CGFloat]) -> CGFloat {
-    guard numbers.count > 0 else { return 0 }
-    return numbers.reduce(0, combine: +) / CGFloat(numbers.count)
-}
-
 public func circumferenceForRadius(radius: CGFloat) -> CGFloat {
     return radius * CGFloat(M_PI * 2)
 }
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -81,8 +81,7 @@ extension SequenceType where Generator.Element: Equatable {
                 self.ip_countOfElement($0) > self.ip_countOfElement($1)
             }
         guard let first = sortedUniqueElements.first else { return nil }
-        print(sortedUniqueElements)
-        return sortedUniqueElements.filter {
+        return sortedUniqueElements.lazy.filter {
             self.ip_countOfElement(first) == self.ip_countOfElement($0)
         }
     }
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -60,3 +60,43 @@ extension Array {
         }
     }
 }
+
+extension CollectionType {
+    public var ip_middleValues: [Generator.Element]? {
+        guard count > 0 else { return nil }
+        let needsAverageOfTwo = self.count.toIntMax().ip_isEven
+        let middle = startIndex.advancedBy(self.count / 2)
+        if needsAverageOfTwo {
+            let leftOfMiddle = startIndex.advancedBy((self.count / 2) - 1)
+            return [self[middle], self[leftOfMiddle]]
+        } else {
+            return [self[middle]]
+        }
+    }
+}
+
+extension CollectionType where Generator.Element: Equatable {
+    public func ip_mostCommonElements() -> [Generator.Element]? {
+        guard let first = self.first else { return nil }
+        let uniqueElements = self.ip_uniqueValues().sort {
+                self.ip_countOfElement($0) > self.ip_countOfElement($1)
+            }
+        return uniqueElements.filter {
+            self.ip_countOfElement(first) == self.ip_countOfElement($0)
+        }
+    }
+        
+    public func ip_uniqueValues() -> [Generator.Element] {
+        var buffer: [Generator.Element] = []
+        for element in self {
+            if !buffer.contains(element) {
+                buffer.append(element)
+            }
+        }
+        return buffer
+    }
+    
+    public func ip_countOfElement(element: Generator.Element) -> Int {
+        return self.filter { $0 == element }.count
+    }
+}
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -64,10 +64,10 @@ extension Array {
 extension CollectionType {
     public var ip_middleValues: [Generator.Element]? {
         guard count > 0 else { return nil }
-        let needsAverageOfTwo = self.count.toIntMax().ip_isEven
-        let middle = startIndex.advancedBy(self.count / 2)
+        let needsAverageOfTwo = count.toIntMax().ip_isEven
+        let middle = startIndex.advancedBy(count / 2)
         if needsAverageOfTwo {
-            let leftOfMiddle = startIndex.advancedBy((self.count / 2) - 1)
+            let leftOfMiddle = startIndex.advancedBy((count / 2) - 1)
             return [self[middle], self[leftOfMiddle]]
         } else {
             return [self[middle]]
@@ -75,7 +75,7 @@ extension CollectionType {
     }
 }
 
-extension CollectionType where Generator.Element: Equatable {
+extension SequenceType where Generator.Element: Equatable {
     public func ip_mostCommonElements() -> [Generator.Element]? {
         let sortedUniqueElements = self.ip_uniqueValues().sort {
                 self.ip_countOfElement($0) > self.ip_countOfElement($1)
@@ -89,7 +89,7 @@ extension CollectionType where Generator.Element: Equatable {
         
     public func ip_uniqueValues() -> [Generator.Element] {
         var buffer: [Generator.Element] = []
-        for element in self {
+        forEach { element in
             if !buffer.contains(element) {
                 buffer.append(element)
             }
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -77,11 +77,12 @@ extension CollectionType {
 
 extension CollectionType where Generator.Element: Equatable {
     public func ip_mostCommonElements() -> [Generator.Element]? {
-        guard let first = self.first else { return nil }
-        let uniqueElements = self.ip_uniqueValues().sort {
+        let sortedUniqueElements = self.ip_uniqueValues().sort {
                 self.ip_countOfElement($0) > self.ip_countOfElement($1)
             }
-        return uniqueElements.filter {
+        guard let first = sortedUniqueElements.first else { return nil }
+        print(sortedUniqueElements)
+        return sortedUniqueElements.filter {
             self.ip_countOfElement(first) == self.ip_countOfElement($0)
         }
     }
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -62,7 +62,10 @@ extension Array {
 }
 
 extension CollectionType {
-    public var ip_middleValues: [Generator.Element] {
+    /// This grabs the element(s) in the middle of the array without doing any sorting.
+    /// If there's an odd number the return array is just one element.
+    /// If there are an even number it will return the two middle elements.
+    public var ip_middleElements: [Generator.Element] {
         guard count > 0 else { return [] }
         let needsAverageOfTwo = count.toIntMax().ip_isEven
         let middle = startIndex.advancedBy(count / 2)
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -62,8 +62,8 @@ extension Array {
 }
 
 extension CollectionType {
-    public var ip_middleValues: [Generator.Element]? {
-        guard count > 0 else { return nil }
+    public var ip_middleValues: [Generator.Element] {
+        guard count > 0 else { return [] }
         let needsAverageOfTwo = count.toIntMax().ip_isEven
         let middle = startIndex.advancedBy(count / 2)
         if needsAverageOfTwo {
@@ -76,13 +76,13 @@ extension CollectionType {
 }
 
 extension SequenceType where Generator.Element: Equatable {
-    public func ip_mostCommonElements() -> [Generator.Element]? {
+    public func ip_mostCommonElements() -> [Generator.Element] {
         let sortedUniqueElements = self.ip_uniqueValues().sort {
-                self.ip_countOfElement($0) > self.ip_countOfElement($1)
+                self.ip_countOf($0) > self.ip_countOf($1)
             }
-        guard let first = sortedUniqueElements.first else { return nil }
+        guard let first = sortedUniqueElements.first else { return [] }
         return sortedUniqueElements.lazy.filter {
-            self.ip_countOfElement(first) == self.ip_countOfElement($0)
+            self.ip_countOf(first) == self.ip_countOf($0)
         }
     }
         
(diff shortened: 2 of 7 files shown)


SwiftWisdom/Core/Geometry/Geometry.swift
```diff
@@ -86,6 +86,11 @@ public func random(range: Range<Int>) -> Int {
 }
 
 public func avg(numbers: CGFloat...) -> CGFloat {
+    return avg(numbers)
+}
+
+public func avg(numbers: [CGFloat]) -> CGFloat {
```
loganwright: I think there's some value to having `avg` as a global function as well, but it might be better to also move it to a generic implementation for greater flexibility and usefulness.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -54,3 +54,43 @@ extension Array {
...
+        forEach { element in
+            if !buffer.contains(element) {
+                buffer.append(element)
+            }
+        }
+        return buffer
+    }
+    
+    public func ip_countOfElement(element: Generator.Element) -> Int {
+        return self.filter { $0 == element }.count
```
loganwright: Space or newline after `}`

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -54,3 +54,43 @@ extension Array {
         }
     }
 }
+
+extension CollectionType {
+    public var ip_middleValues: [Generator.Element]? {
```
LoganWright: I think in its current implementation, the use of this function is kind of unclear.  It appears to be getting the middle values of the collection, but kind of sounds like it might use comparable to get middle.  I'm also not sure about the array return.  Let's pull this one out while we think about it.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -54,3 +54,43 @@ extension Array {
...
+            let leftOfMiddle = startIndex.advancedBy((count / 2) - 1)
+            return [self[middle], self[leftOfMiddle]]
+        } else {
+            return [self[middle]]
+        }
+    }
+}
+
+extension SequenceType where Generator.Element: Equatable {
+    public func ip_mostCommonElements() -> [Generator.Element]? {
```
loganwright: I think this one could also be more thoroughly defined, some of the behavior will be unexpected.  ie: `[1,2,3,4,5]` would return the first element. I think some of these should be moved to another PR so we can get the sure things merged in

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -54,3 +54,43 @@ extension Array {
...
+        let sortedUniqueElements = self.ip_uniqueValues().sort {
+                self.ip_countOfElement($0) > self.ip_countOfElement($1)
+            }
+        guard let first = sortedUniqueElements.first else { return nil }
+        return sortedUniqueElements.lazy.filter {
+            self.ip_countOfElement(first) == self.ip_countOfElement($0)
+        }
+    }
+        
+    public func ip_uniqueValues() -> [Generator.Element] {
```
loganwright, line 92: :+1:

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
paulrolfe: That would return the original array if they all have a count of 1, actually. Some of the average functions are dependent on this, so I say let's figure it out now.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
paulrolfe: Other average functions depend on this one, so we could make it internal or something while we think about it. I'd prefer to figure out a way to have this make sense. This does not use comparability, but I don't know how to say that.

loganwright: Looking forward, I see you're using this in mean, so maybe it has some value, but let's tweak the api a bit to return a non-optional array, and add some spacing / cleanup.  Mostly just saving count so it doesn't run multiple times, newlining array

``` Swift
    func ip_medianValues() -> [Generator.Element] {
        guard !isEmpty else { return [] }
        let count = self.count
        let half = count / 2
        let center = left.advancedBy(1)
        if count.ip_isEven {
            let left = startIndex.advancedBy(half - 1)
            return [
                self[left],
                self[center]
            ]
        } else {
            return [
                self[center]
            ]
        }
    }
```

How do you feel about `medianValues`?

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -54,3 +54,43 @@ extension Array {
...
+        var buffer: [Generator.Element] = []
+        forEach { element in
+            if !buffer.contains(element) {
+                buffer.append(element)
+            }
+        }
+        return buffer
+    }
+    
+    public func ip_countOfElement(element: Generator.Element) -> Int {
```
loganwright: I think this should be `ip_countOf` to have better parity w/ swift apis, ie: `indexOf`

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,59 @@
...
+}
+
+extension CollectionType where Generator.Element: Mathable {
+    
+    public func ip_average() -> Generator.Element? {
+        return ip_mean()
+    }
+    
+    public func ip_mean() -> Generator.Element? {
+        let zero = Generator.Element(0.toIntMax())
```
loganwright: We use zero a lot, it might be worth writing an extension on mathable:

``` Swift
extension Mathable {
    public static var zero: Self {
        return self.init(0.toIntMax())
    }
}
```

Then here, it'd be:

``` Swift
let zero = Generator.Element.zero
```

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+    
+    public func ip_median() -> Generator.Element? {
+        let zero = Generator.Element(0.toIntMax())
+        let count = Generator.Element(self.count.toIntMax())
+        guard count > zero else { return nil }
+        return sort(<).ip_middleValues?.ip_mean()
+    }
+    
+    public func ip_mode() -> [Generator.Element]? {
```
loganwright: I think this should return non-optional array

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,59 @@
...
+        return ip_mostCommonElements()
```
loganwright, line 42: I think we can pull it out and just leave mode for now, but there needs to be a check for no repeated values.  I believe if no value is repeated or takes precedence, then there is no mode.  It might make sense to do this with a dictionary and constrain to hashable to make it easier.  We should only be using numbers anyway so hashable should be easy to come by

SwiftWisdom/Core/Geometry/Geometry.swift
paulrolfe: Will do, I'll make it an optional return as well

loganwright:
Some clarification on a few things for me, then I'm good.  Let's get a second opinion on some of these though, math isn't my strong suit.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
paulrolfe: My issue with this is that median does imply that it uses comparison, but we don't do any sorting -- we are really just grabbing the middle values. I made this separate because I thought grabbing the middle elements from an array would have some use outside of determining the median.

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
paulrolfe: I like that

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
paulrolfe, line 42: I asked some folks around my desk when writing this and came to the conclusion we should return each element when no value(s) take precedence -- let the user deal with it however they want. I'm curious what other people think though.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
paulrolfe: I dig it

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
loganwright: Let's field some external opinions on this one as well to see if we can get some opinions one way or the other.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
loganwright: That sounds ok, let's make it non-optional return

SwiftWisdom/Core/Geometry/Geometry.swift
loganwright: Is there a reason to have it optional? I'd rather see it return 0 personally.  Again, I might be the minority on this one :smile:

paulrolfe: I'm just going with what @cprime said. I can see both sides of this: If you've got 0 values, there is no average. But also if you want an average of nothing, that would be 0, right?

I just think it should be consistent between this and the `ip_mean()` function.

loganwright: I prefer not to deal w/ optionals in collections and numbers for the most part.  If the array is empty and shouldn't average, the user can check it before running average. @cprime thoughts?

djbrooks111: The count of an empty array, which in essence is "no values" is always returned as a non-optional 0.  I think we should follow suit

paulrolfe: Boom. Settled.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -54,3 +54,43 @@ extension Array {
         }
     }
 }
+
+extension CollectionType {
+    public var ip_middleValues: [Generator.Element] {
```
brightredchilli: This seems like a strange function...what is the use of middleValues?

Benuuu: @brightredchilli this calculates median.  It should probably be called ip_median

paulrolfe: I use it in finding the median (after sorting an array), but I made it public because I imagined it might be nice to have an easy way to find the element(s) at the middle of an array.

Benuuu: Actually that's wrong because it doesn't sort.

Benuuu: :+1: what's the use of middleValues?

paulrolfe: @Benuuu @brightredchilli see my comment above -- there's a different function for median that sorts the array and then grabs the middle value using this.

You aren't the first people to be confused by it, but not sure what the fix is. I think it's valuable to be able to get the middle element(s) from an array though.

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,73 @@
...
+    func + (lhs: Self, rhs: Self) -> Self
+    func / (lhs: Self, rhs: Self) -> Self
+    func * (lhs: Self, rhs: Self) -> Self
+    func - (lhs: Self, rhs: Self) -> Self
+    init(_ int: IntMax)
+}
+
+extension Mathable {
+    public static var zero: Self {
+        return self.init(0.toIntMax())
```
brightredchilli, line 22: what is toIntMax for here?

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+extension CollectionType where Generator.Element: Mathable {
+    
+    public func ip_average() -> Generator.Element {
```
brightredchilli: I feel like two functions that does the same thing is overkill - maybe stick with mean?

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+    
+    public func ip_median() -> Generator.Element {
+        let zero = Generator.Element.zero
+        let count = Generator.Element(self.count.toIntMax())
+        guard count > zero else { return zero }
+        return sort(<).ip_middleValues.ip_mean()
+    }
+    
+    public func ip_mode() -> [Generator.Element] {
```
brightredchilli, line 41: same comment here

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -0,0 +1,73 @@
...
+//  SwiftWisdom
+//
+//  Created by Paul Rolfe on 2/12/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+public protocol Mathable: Comparable {
```
brightredchilli, line 12: in researching this I realized that there is a protocol called IntegerArithmeticType.
https://developer.apple.com/library/tvos/documentation/Swift/Reference/Swift_IntegerArithmeticType_Protocol/index.html#//apple_ref/swift/intfm/IntegerArithmeticType/s:FPSs21IntegerArithmeticType8toIntMaxuRq_S__Fq_FT_VSs5Int64

this implements the operators below. Should Mathable simply inherit from this protocol?

SwiftWisdomTests/CollectionTypeTests.swift
```diff
@@ -0,0 +1,27 @@
...
+        let strings = ["b", "b", "v", "v"]
+        XCTAssertEqual(strings.ip_mostCommonElements(), ["b", "v"])
+        let numbers = [1, 2, 3, 4, 5, 6, 6, 7]
+        let noMode = [1, 2, 3, 4, 5]
+        var empty = numbers
+        empty.removeAll()
+        XCTAssertEqual(numbers.ip_mostCommonElements(), [6])
+        XCTAssertEqual(noMode.ip_mostCommonElements(), noMode)
+        XCTAssertEqual(empty.ip_mostCommonElements(), [])
+
```
brightredchilli: empty line here

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
paulrolfe, line 22: Because `IntMax` is the type for the Mathable initializer. I'm pretty sure we use that because all number types can initialize from `IntMax` (aka Int64) -- but @LoganWright actually came up with using that initializer for Mathables.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
brightredchilli: I understand middle, but not middleValues. Maybe some documentation would help.

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
paulrolfe: I'm down with that. There's always the global `func avg(...` if you really want to use the word average.

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
paulrolfe, line 41: I like the symmetry of having mean, median, and mode for these mathable types. And then if your array / collection isn't mathable then you can still find the `ip_mostCommonElements`. It would feel incomplete with just mean and median.

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
paulrolfe, line 12: That would exclude floats and doubles.

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
loganwright, line 22: Yea, the reason we use `IntMax` is because almost all primitive numeric types can take it as an initializer.  My guess is that under the covers, this is how things like this work:

```
let a: Double = 0
```

etc.

SwiftWisdomTests/MathableTests.swift
```diff
@@ -0,0 +1,70 @@
...
+    func testMean() {
+        let maths = [4, 5, 6, 9, 2, 4]
+        let cgfloatmaths = maths.map { CGFloat($0) }
+        let floatMaths = maths.map { Float($0) }
+        XCTAssertEqual(maths.ip_mean(), 5)
+        XCTAssertEqual(cgfloatmaths.ip_mean(), 5)
+        XCTAssertEqual(floatMaths.ip_mean(), 5)
+        var none = maths
+        none.removeAll()
+        XCTAssertEqual(none.ip_average(), 0)
```
brightredchilli: does this now throw an error? we removed ip_average, yes?

brightredchilli:
:ship: for me, after making sure tests still run

PR #23, opened 2016-02-16
public initializer

loganwright:


paulrolfe:
:ship:

brightredchilli:
What is the tag naming convention -b for?

paulrolfe:
Beta, I presume. @LoganWright should bump that and re-tag before merge.

brightredchilli:
Right, but it's all beta. Don't see how this is different than a tag bump. I'm not resisting, just trying to understand why there is a difference from incrementing a number, so that we can document this.

loganwright:
@brightredchilli I use it for the situation where I'm using changes that aren't yet approved and merged into master, so this way if someone has `~> 0.1.0` in their podspec, they don't update until the code's been approved. 

Also allows us to bump from `0.1.1-beta` to `0.1.1` so we don't have a ton of increments.

I'm fine just bumping versions, but I think the `-b` is good practice for using a pod pre merge to prevent others from pulling it unwillingly.

PR #25, opened 2016-02-21
adding umbrella podspec

loganwright:
I'd like to propose a new method of managing our pods, as of now, our two main collections are:

```
import IntrepidSwiftWisdom
import IP_UIKit_Wisdom
```

A lot of this is due to name spacing, but if we're going to make the decision to be public with these, and host them on the public repository, I think one simple `Intrepid` import will be nicer.  Once we own the namespace on cocoapods, as long as that is our choice of import, there will not be conflicts.

```
import Intrepid
```

Where it should live:

I'm proposing it live in swift wisdom for now, but potentially break out into its own library in the future

How do we control sub libraries:

We may be able to leverage subspecs, ie:

```
pod 'Intrepid/UIKit'
pod 'Intrepid/Video'
```

or just

```
pod 'Intrepid'
```

to default to more comprehensive coverage.

We might be able to leverage the new `@_exported import` to include sub libraries in our overall header.  ie: `import Intrepid` could import Swift Wisdom and UIKit Wisdom w/o requiring an explicit import.
## Issues

I think the main potential issue is bloat, but as of now, I don't see it as a problem to include a few extra unused files.  This could also be managed through subspecs for developers to opt in only to the classes they want.

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,0 +1,22 @@
+Pod::Spec.new do |s|
+  s.name          = "Intrepid"
+  s.version       = "0.1.3"
+  s.summary       = "Swift Bag"
+  s.description   = <<-DESC
+                    Collection of extensions and utility classes by and for the developers at intrepid pursuits.
+                    DESC
+  s.homepage      = "https://github.com/IntrepidPursuits/swift-wisdom"
+  s.license       = "MIT"
+  s.authors       = { "Intrepid Pursuits" => "logan@intrepid.io" }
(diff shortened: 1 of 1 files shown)


brightredchilli:
I dig this, :ship: for me. We should already be breaking down the files by 'category'. So, if I'm reading this right, we should be able to add another podspec file for Intrepid/UIKit to just include UIKit.

loganwright:
@brightredchilli You are reading it correctly! Except it wouldn't be another podspec file, just a subspec added to this one.

PR #29, opened 2016-03-16
gradient crash fix

paulrolfe:


Benuuu:
:shipit:

loganwright:
:shipit: (bump podspec, tag etc.)

Benuuu:
@LoganWright the bump podspec etc be independent of the PR right?

loganwright:
@Benuuu We have been bumping/tagging it on the PR, then merging it in together. Any arguments against that pattern?

PR #30, opened 2016-03-21
fixing comparison bug

loganwright:


SwiftWisdomTests/VersionTests.swift
```diff
@@ -35,4 +35,16 @@ class VersionTests: XCTestCase {
         
         XCTAssert(highest > middle)
     }
+    
+    func testVersionLessThan() {
+        let left = Version(string: "1.0.0")!
+        let right = Version(string: "0.5.2")!
+        XCTAssertFalse(left < right)
+    }
```
Benuuu, line 51: Could you write some more tests for this that improve the coverage?

loganwright, line 51: @Benuuu Sure, any suggestions on what I should add?

paulrolfe, line 51: Maybe something with no patch? One with a revision? One with gibberish in the initializer string? `33,64/f4`

paulrolfe:
:shipit:

SwiftWisdomTests/VersionTests.swift
loganwright, line 51: Boom, done!

PR #31, opened 2016-03-23
Ying/changes

brightredchilli:


SwiftWisdom/Core/UIKit/UICollectionView+FlowLayout.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+//
+
+import Foundation
+
+public extension UICollectionView {
+    var flowLayout: UICollectionViewFlowLayout? {
+        return self.collectionViewLayout as? UICollectionViewFlowLayout
+    }
+}
```
Benuuu: newline

Benuuu:
:shipit:

loganwright:
Looks good to me as well after Ben's comment. :shipit:

loganwright:
Ying, can you also update the podspec and tag it? Do you have a trunk account? If you send me your email, I'll add you to auth list for pushing.

brightredchilli:
i will do so. it also appears that...we have a code convention of namespacing extensions...do you guys think I should do that here?

PR #32, opened 2016-03-25
fix xcode 7.3 warnings

brightredchilli:


brightredchilli:
The podspec bump commit(61667ce) appears to be spurious, same SHA is found on IntrepidPursuits/swift-wisdom.

For these changes, I propose we bump the minor number, although we really should bump the minor number to give us some maneuvering room to have backwards compat to the previous selector syntax

SwiftWisdom/Core/NSData/NSData+Extensions.swift
```diff
@@ -20,8 +20,10 @@ public extension NSData {
         bufer.forEach {
             let hi  = Int(($0 & 0xf0) >> 4)
             let low = Int($0 & 0x0f)
-            output[ix++] = hexChars[ hi]
-            output[ix++] = hexChars[low]
+            ix += 1
```
loganwright: This isn't quite the right translation, it should be

``` Swift
output[ix] = hexChars[hi]
ix += 1
output[ix] = hexChars[low]
ix += 1
```

For example:

``` Swift
var a = 1
var arr = [1,2,3,4]
arr[a++]
```

Returns `2`, not `3`

I think this is part of the reason they deprecated syntax :smile:

SwiftWisdom/Core/NSData/NSData+Extensions.swift
```diff
@@ -116,10 +118,10 @@ extension NSData {
 
 extension NSData {
     public func ip_segmentGenerator(start start: Int = 0, chunkLength: Int) -> AnyGenerator<NSData> {
-        guard let segmentToWrite = ip_suffixFrom(start) else { return anyGenerator { return nil } }
+        guard let segmentToWrite = ip_suffixFrom(start) else { return AnyGenerator { return nil } }
```
loganwright, line 121: I think maybe newline here, there's a lot of `{}`

loganwright:
@brightredchilli Looks good to me, a couple of changes around the `++`, then good to :shipit: 

This feels like a minor bump to me as well, if that's what you're saying in the last sentence. I think one of the `minor number`s was meant to be something else in your comment, so let me know if I missed something.

brightredchilli:
thanks, really good catch on the prefix operator! No, yeah I think I will bump to 0.2.0

PR #34, opened 2016-03-29
Email validation should also expect a character after '.'

Benuuu:


loganwright:
:ship:

SwiftWisdom/Core/StandardLibrary/String/String+EmailValidation.swift
```diff
@@ -17,6 +17,6 @@ extension String {
      - returns: True if the string contains an @ followed by a . with characters before between and after.
      */
     public func ip_isValidEmail() -> Bool {
-        return rangeOfString(".@.+\\.", options: .RegularExpressionSearch) != nil
+        return rangeOfString(".@.+\\..", options: .RegularExpressionSearch) != nil
```
paulrolfe, line 20: I'm no regex expert, but isn't there an extra `\` here? It looks like it's escaping a `\` instead of the `.`

Benuuu, line 20: It's because I need to escape the `\` in a swift string.  The regex is really `.@.+\..`

paulrolfe, line 20: oh right! thanks

paulrolfe:
:shipit:

PR #35, opened 2016-03-30
fixing overload causing ambiguous use of cell dequeing

loganwright:
The overload was making the default arguments behave weird and would cause `ambiguous use`

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.2.1"
+  s.version       = "0.2.2"
   s.summary       = "Swift Bag"
   s.description   = <<-DESC
                     Collection of extensions and utility classes by and for the developers at intrepid pursuits.
--- a/SwiftWisdom/Core/UIKit/CellConfiguring.swift
+++ b/SwiftWisdom/Core/UIKit/CellConfiguring.swift
@@ -31,11 +31,7 @@ public extension UICollectionView {
     
     // MARK: Register
 
-    public func ip_registerCell<T where T: UICollectionViewCell, T: TypeIdentifiable>(type: T.Type) {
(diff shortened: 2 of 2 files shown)


brightredchilli:
i'm using the collection cell overload without issue - what error are you getting?

loganwright:
@brightredchilli the collection cell overload works because it's declared like:

``` Swift
func registerCell(type: Type) { ...
func registerCell(type: Type, identifier: String) { ...
```

But the tableview cell one is declared like:

``` Swift
func registerCell(type: Type) { ...
func registerCell(type: Type, identifier: String = default) { ...
```

So if you call `tv.registerCell(someType)` it can't tell if you want the overload, or the default and gives ambiguous error. I think we should opt to just use default argument in both.

LoganWright:
@brightredchilli the collection cell overload works because it's declared like:

```Swift
func registerCell(type: Type) { ...
func registerCell(type: Type, identifier: String) { ...
```

But the tableview cell one is declared like:

```Swift
func registerCell(type: Type) { ...
func registerCell(type: Type, identifier: String = default) { ...
```

So if you call `tv.registerCell(someType)` it can't tell if you want the overload, or the default and gives ambiguous error. I think we should opt to just use default argument in both.

vinnyt:
:+1:

SwiftWisdomTests/NSDataSubdataTests.swift
```diff
@@ -100,6 +100,8 @@ class NSDataSubdataTests: XCTestCase {
         let combination = [zero, one, two, three].joinWithSeparator("")
         let data = combination.ip_dataFromHexadecimalString()!
         
+        
+
```
intrepidmatt: Extraneous whitespace...did you mean to commit changes to this file?

brightredchilli:
:ship:

brightredchilli:
my apologies, was too gung ho with this and never tested the uitableview version

PR #37, opened 2016-03-31
Lw/rx

loganwright:
@Benuuu @vinnyt 

I got these from your project, if there's any more that could be added, that'd be awesome!

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.2.2"
+  s.version       = "0.2.3"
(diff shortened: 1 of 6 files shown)


vinnyt:
:+1: :+1:

Benuuu:
:shipit: 

Not sure if we need / want examples somewhere.

vinnyt:
We should consider cooking up some examples

loganwright:
@Benuuu @vinnyt I don't feel qualified to write examples yet, if one of you guys wouldn't mind adding some, that would be :rocket:!

vinnyt:
I can take a cut at this in a couple of weeks  unless @Benuuu can get to it sooner.

PR #40, opened 2016-04-15
Pat/typography

ButkiewiczP:
Adds an extension to UIILabel for typography related functions.

Two convenience methods are exposed in `UILabel+Typography.swift`: `ip_setCharacterSpacing` and `ip_setLineSpacing`.

Both methods determine the current text of the label, either pulling the raw text, or in the case of an attributed label, pulling the text and attributes. It then applies another attribute (depending on the method) across the range of the string. This prevents the methods from overwriting other attributed changes that may have been made to the label.

I struggled with whether or not this should have been an extension on `NSAttributedString` or on `UILabel`. In real world use, I found being able to call `myLabel.ip_setCharacterSpacing(4.0)` to be the easiest way.

### DIFF
--- a/SwiftWisdom/Core/UIKit/UILabel+Typography.swift
+++ b/SwiftWisdom/Core/UIKit/UILabel+Typography.swift
@@ -1,0 +1,58 @@
+//
+//  UILabel+Typography.swift
+//  SwiftWisdom
+//
+//  Created by Patrick Butkiewicz on 4/15/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+extension UILabel {
+    
+    /**
+     Sets the text of the UILabel instance to an attributed string with tracking applied
+     - Parameter value: The value in points that each character is kerned to
+     */
+    
+    public func ip_setCharacterSpacing(value: CGFloat) {
+        addAttribute(NSKernAttributeName, value: value)
+    }
+    
+    /**
+     Sets the text of the UILabel instance to an attributed string with the specified line spacing
+     - Parameter value: The space, in points, between each line.
+     */
+    
+    public func ip_setLineSpacing(value: CGFloat) {
+        let paragraphSpacing = NSMutableParagraphStyle()
+        paragraphSpacing.lineSpacing = value
+        paragraphSpacing.alignment = textAlignment
+        addAttribute(NSParagraphStyleAttributeName, value: paragraphSpacing)
+    }
+    
+    // MARK: Private
+    
+    private func addAttribute(attr: String, value: AnyObject) {
+        let attrText = mutableAttributedText()
+        attrText.addAttributes([attr:value], range: NSMakeRange(0, attrText.length))
+        text = nil
+        attributedText = attrText
+    }
+    
+    private func baseAttributes() -> [String:AnyObject] {
+        return [NSFontAttributeName : font, NSForegroundColorAttributeName : textColor]
+    }
+    
+    private func mutableAttributedText() -> NSMutableAttributedString {
+        let bareText = text ?? ""
+        
+        if let attributedText = attributedText {
+            return NSMutableAttributedString(attributedString: attributedText)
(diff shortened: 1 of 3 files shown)


SwiftWisdom/Core/UIKit/UILabel+Typography.swift
```diff
@@ -0,0 +1,58 @@
...
+    private func baseAttributes() -> [String:AnyObject] {
+        return [NSFontAttributeName : self.font, NSForegroundColorAttributeName : self.textColor]
+    }
+    
+    private func mutableAttributedText() -> NSMutableAttributedString {
+        let bareText = self.text ?? ""
+        let attributedText = self.attributedText
+        
+        if let attributedText = attributedText {
+            return attributedText.mutableCopy() as! NSMutableAttributedString
```
loganwright: Would it be easier to do `NSMutableAttributedString(attributedString: attributedText)`

vinnyt:
+1 from my perspective

loganwright:
:shipit: Do you have trunk access to cocoapods?

SwiftWisdom/Core/UIKit/UILabel+Typography.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    private func addAttribute(attr: String, value: AnyObject) {
+        let attributedText = mutableAttributedText()
+        attributedText.addAttributes([attr:value], range: NSMakeRange(0, attributedText.length))
+        self.text = nil
+        self.attributedText = attributedText
+    }
+    
+    private func baseAttributes() -> [String:AnyObject] {
+        return [NSFontAttributeName : self.font, NSForegroundColorAttributeName : self.textColor]
```
paulrolfe: This leaves out alignment, right? Maybe make an NSParagraphStyle to capture that?

```
let newParagraphStyle = NSMutableParagraphStyle()
newParagraphStyle.alignment = textAlignment
return [
   NSFontAttributeName : self.font, 
   NSForegroundColorAttributeName : self.textColor, 
   NSParagraphStyleAttributeName : newParagraphStyle
]
```

SwiftWisdom/Core/UIKit/UILabel+Typography.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    /**
+     Sets the text of the UILabel instance to an attributed string with the specified line spacing
+     - Parameter value: The space, in points, between each line.
+     */
+    
+    public func ip_setLineSpacing(value: CGFloat) {
+        let paragraphSpacing = NSMutableParagraphStyle()
+        paragraphSpacing.lineSpacing = value
+        addAttribute(NSParagraphStyleAttributeName, value: paragraphSpacing)
```
paulrolfe, line 32: This could override some existing paragraph style? Can you try to grab that and edit it before assigning a new one?

SwiftWisdom/Core/UIKit/UILabel+Typography.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+    
+    // MARK: Private
+    
+    private func addAttribute(attr: String, value: AnyObject) {
+        let attributedText = mutableAttributedText()
+        attributedText.addAttributes([attr:value], range: NSMakeRange(0, attributedText.length))
+        self.text = nil
```
paulrolfe: we don't usually use `self` unless we have to

paulrolfe:
:shipit:

SwiftWisdom/Core/UIKit/UILabel+Typography.swift
ButkiewiczP, line 32: I'm not sure it would override existing paragraph styles, unless they were also defined on the whole string. Theoretically, this should just add another attribute around the entirety of the string. I'm not sure how it would affect conflicting paragraph style properties though. The other option would be to enumerate all attributes, identify paragraph styles and set their line spacing attribute.

PR #41, opened 2016-04-23
Added ip_toMagnitudeString method to Int+Extensions

wingchi:
`ip_toMagnitudeString` converts an Integer Type to a String with a rounded value and a magnitude suffix.

e.g. - for one decimal place 1234 becomes "1.2K", 5678000 becomes "5.7M", etc.

One decimal place is the default.

This method is useful for formatting a number when it needs to be displayed in a small space and the exact value isn't necessary, such as the number likes on a Facebook post.


SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
         (0..<self).forEach { _ in closure() }
     }
 }
+
+extension IntegerType {
+    public func ip_toMagnitudeString(decimalPlaces decimals: Double = 1) -> String {
+        
+        print(self)
```
loganwright: let's remove all print statements

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
         (0..<self).forEach { _ in closure() }
     }
 }
+
+extension IntegerType {
+    public func ip_toMagnitudeString(decimalPlaces decimals: Double = 1) -> String {
```
loganwright: I think the signature should take an `Int`, I don't see anyone using `1.5` decimal places. I also think `decimalPlaces` should be used internally and externally. `decimals` reads a bit like the number of `.` instead of the number of places

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
...
+        }
+        
+        let formatter = NSNumberFormatter()
+        formatter.numberStyle = .DecimalStyle
+        formatter.minimumFractionDigits = 0
+        formatter.maximumFractionDigits = Int(decimals)
+        
+        guard let likesFormatted = formatter.stringFromNumber(roundedLikes) else { return "\(roundedLikes)\(units[magnitude-1])" }
+        
+        return "\(likesFormatted)\(units[magnitude-1])";
```
loganwright: Should we access more safely? It seems feasible to receive a number in the zillions which would crash on index out of bounds.

loganwright: no `;`

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
...
+            roundedLikes /= 1000.0
+            magnitude += 1
+        }
+        
+        let formatter = NSNumberFormatter()
+        formatter.numberStyle = .DecimalStyle
+        formatter.minimumFractionDigits = 0
+        formatter.maximumFractionDigits = Int(decimals)
+        
+        guard let likesFormatted = formatter.stringFromNumber(roundedLikes) else { return "\(roundedLikes)\(units[magnitude-1])" }
```
loganwright: newline the contents of the `else`

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
...
+        // 7563 / 10 ^ 2 = 75.63
+        var roundedLikes = round(pow(10.0, decimals) * likes / pow(1000.0, Double(magnitude))) / pow(10.0, decimals)
+        
+        // if rounding brings our display number over 1000, divide by 1000 and then bump the magnitude
+        if roundedLikes >= 1000 {
+            roundedLikes /= 1000.0
+            magnitude += 1
+        }
+        
+        let formatter = NSNumberFormatter()
```
loganwright: I think this section through to `likesFormatted` where we're rounding the nearest 2 digit decimal place could be a separate function.

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
...
+        var magnitude: Int = Int(log10(likes) / 3.0) // the order of magnitude of our like count in thousands
+        
+        // this formula rounds the likes to the nearest 1000s magnitude with X decimal points
+        // E.G. - 75633 becomes 75.63
+        // if X is 2
+        // pow(1000.0, Double(magnitude)) = 1000^1 = 1000
+        // 10 ^ 2 * likes / 1000 = 7563.3
+        // round(7563.3) = 7563
+        // 7563 / 10 ^ 2 = 75.63
+        var roundedLikes = round(pow(10.0, decimals) * likes / pow(1000.0, Double(magnitude))) / pow(10.0, decimals)
```
loganwright: break this out, too complex for a single line

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,41 @@ extension IntegerType {
...
+        
+        // if rounding brings our display number over 1000, divide by 1000 and then bump the magnitude
+        if roundedLikes >= 1000 {
```
loganwright: Is there a chance it will be over 2000 units? Would something like this be more applicable:

``` Swift
while roundedLikes >= 1000 {
    roundedLikes /= 1000
    magnitude += 1
}
```

SwiftWisdomTests/IntegerTypeTests.swift
```diff
@@ -24,4 +24,20 @@ class IntegerTypeTests: XCTestCase {
         count.ip_times { collection.append("Hi") }
         XCTAssert(count == collection.count)
     }
+    
+    func testToMagnitudeString() {
```
loganwright, line 28: 👍

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,34 @@ extension IntegerType {
...
+        
+        // if rounding brings our display value over 1000, divide by 1000 and then bump the magnitude
+        if roundedHundredsValue >= 1000 {
+            roundedHundredsValue /= 1000.0
+            magnitude += 1
+        }
+        
+        let formatter = DecimalFormatter.sharedInstance
+        guard let likesFormatted = formatter.stringFromNumber(roundedHundredsValue) else {
+            return "\(roundedHundredsValue)\(units[magnitude-1])"
```
loganwright: I'm concerned here as well about crash potential. I know it's rare, but I think we can check that this index is within range first and offer some type of alternative to crash. 

There might be an `ip_safe` call, not sure if it's been added and we could do: `let suffix = units[ip_safely: magnitude - 1] ?? alt`. There's some implications here where the value won't be correct, so maybe someone more familiar w/ formatting can offer a solution.

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -24,3 +24,34 @@ extension IntegerType {
...
+        guard self > 999 else { return "\(self)" }
+        
+        let units = ["K", "M", "B", "T", "Q"]
+        
+        let value = Double(self.toIntMax())
+        
+        var magnitude: Int = Int(log10(value) / 3.0) // the order of magnitude of our like count in thousands
+        
+        // divide value by 1000^magnitude to get hundreds value, then round to desired decimal places
+        var roundedHundredsValue = (value / pow(1000.0, Double(magnitude))).ip_roundToDecimalPlaces(decimalPlaces)
```
loganwright, line 40: I think breaking out the initial function and then calling round would make this a little easier on the eyes. We have a bit too much logic and parens for a single line

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
loganwright: Did we ever respond to this one, it might never be a case, but wanted to point it out in case it could be 👍

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
wingchi: Didn't know there was `someArray[ip_safe: index]`. Thanks for letting me know.

loganwright:
Looking good to me :shipit: 

Let's get eyes from somebody more mathematically inclined then myself and get this merged! If you haven't worked w/ cocoapods before, let's chat right before you merge and walk through the process of updating the spec 👍

SwiftWisdom/Core/NSFormatter/DecimalFormatter.swift
```diff
@@ -0,0 +1,23 @@
...
+//  DecimalFormatter.swift
+//  SwiftWisdom
+//
+//  Created by Stephen Wong on 5/1/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+
+final class DecimalFormatter: NSNumberFormatter {
```
brightredchilli: im not sure I agree with creating a top level subclass called DecimalFormatter here - after all this does not add any new functionality to NSNumberFormatter. This looks like a convenience initializer or an extension that returns a nsnumberformatter, at most.

SwiftWisdomTests/IntegerTypeTests.swift
```diff
@@ -24,4 +24,20 @@ class IntegerTypeTests: XCTestCase {
         count.ip_times { collection.append("Hi") }
         XCTAssert(count == collection.count)
     }
+    
+    func testToMagnitudeString() {      
+        let numbers = [14, 1234, 15420003, 5678000000, 8888000000000, 999999, 7666555444333222, 1234567891234567890]
+        let formattedNumbersOnePlace = ["14", "1.2K", "15.4M", "5.7B", "8.9T", "1M", "7.7Q", "1.2E18"]
```
brightredchilli, line 30: can we also include "1.0" and "0.001" in this list?

SwiftWisdomTests/DoubleTests.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import XCTest
+import SwiftWisdom
+
+class DoubleTests: XCTestCase {
+
+    func testRounding() {
+        let unroundedDoubles = [M_PI, 5.0156, 10.014, 9999.994]
```
brightredchilli: maybe perhaps also do rounded to 5 decimal places(whereby you should still get the same value back)?

SwiftWisdom/Core/NSFormatter/DecimalFormatter.swift
brightredchilli: Also, I suggest making this private to the file that uses it, unless we feel there is a compelling reason to expose it.

loganwright: +1 to extension that returns number formatter

SwiftWisdomTests/IntegerTypeTests.swift
wingchi, line 30: Only works for `IntegerType`

paulrolfe:
@wingchi Want to add the tests that @brightredchilli mentioned about the 5 decimal places? Then rebase, and :shipit:

PR #42, opened 2016-05-05
add ip_cellForRowAtIndexPath:

brightredchilli:
added another helper method into CellConfiguring

### DIFF
--- a/SwiftWisdom.xcodeproj/project.pbxproj
+++ b/SwiftWisdom.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


Benuuu:
How do all these methods work when you register two cells for the same view?

brightredchilli:
im not sure what you mean, two cells for the same view? They would need to have different identifiers.

Benuuu:
Ah right.  I was confused because you weren't using an identifier in your code, but the other methods call with identifier.

Benuuu:
:shipit:

loganwright:
:shipit:

loganwright:
I think this is good to merge @brightredchilli

loganwright:
closed for manual merge from 0.3.2 tag w/ @Benuuu

brightredchilli:
thanks guys - sorry about this.

PR #43, opened 2016-05-06
Added a method to check if a string contains only whitespace and acco…

nservidio:
…mpanying test.


loganwright:
Looks good to me :shipit:

SwiftWisdom/Core/NSData/String+NSData.swift
```diff
@@ -71,6 +71,10 @@ extension String {
     public var ip_length: Int {
         return self.characters.count
     }
+    
+    public var ip_containsOnlyWhitespace: Bool {
+        return self.stringByTrimmingCharactersInSet(NSCharacterSet.whitespaceCharacterSet()).ip_length == 0
```
wingchi: no newlines? `whitespaceAndNewlineCharacterSet()`

SwiftWisdom/Core/NSData/String+NSData.swift
```diff
@@ -71,6 +71,10 @@ extension String {
     public var ip_length: Int {
         return self.characters.count
     }
+    
+    public var ip_containsOnlyWhitespace: Bool {
+        return self.stringByTrimmingCharactersInSet(NSCharacterSet.whitespaceAndNewlineCharacterSet()).ip_length == 0
+    }
 }
```
Benuuu, line 78: Look at String+Empty.swift

Benuuu:
This also shouldn't exist in String+NSData

Benuuu:
👎

PR #49, opened 2016-07-01
Extensions from Bose

paulrolfe:
Adds extensions for 
- ip_locked
- ip_indicesOf
- ip_removeElements
- ~~ip_minutesAndSecondsString~~
- ip_constrainedBetween

If we don't want anything in here, let me know. ~~I think `minutesAndSecondsString` may be too specific, but we like it. 😄~~


SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -575,12 +587,15 @@
...
 			buildPhases = (
+				21125417B2D0E6F75BB64B48 /* [CP] Check Pods Manifest.lock */,
 				C870EFFC951BFB3182EF5126 /* [CP] Check Pods Manifest.lock */,
 				80232B9B1BF2EE2E00818B6E /* Sources */,
 				80232B9C1BF2EE2E00818B6E /* Frameworks */,
 				80232B9D1BF2EE2E00818B6E /* Resources */,
 				9A1B03B8C9D2B525E066F8AA /* [CP] Embed Pods Frameworks */,
 				D7FE8601F9C058C38C034313 /* [CP] Copy Pods Resources */,
+				7C9A62994AB6B82BDD09F259 /* 📦 Embed Pods Frameworks */,
+				CE9C94B81093D2BE1BA7D4FC /* 📦 Copy Pods Resources */,
```
Benuuu: Did you somehow duplicate build phases?  It also looks like you're using cocoapods 1.0.0.  Could you upgrade to 1.0.1?

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -27,6 +27,22 @@ public extension Array where Element: Equatable {
         }
         return false
     }
+    
+    public mutating func ip_removeElements(elements: [Element]) {
```
Benuuu, line 33: Write tests for these

SwiftWisdom/Core/StandardLibrary/Numbers/Mathable.swift
```diff
@@ -51,6 +51,14 @@ public func avg<T: Mathable>(numbers: [T]) -> T? {
     return numbers.ip_mean()
 }
 
+extension Comparable {
+    public func ip_constrainedBetween(low low: Self, high: Self) -> Self {
```
Benuuu: Instead define this as `ip_constrainBetween` instead of constrained because the latter implies a Bool return.

Benuuu: This method doesn't really belong in Mathable since it has nothing to do with the Mathable protocol

SwiftWisdom.xcodeproj/project.pbxproj
paulrolfe: Yeah I noticed I was on 1.0, after pod install, then upgraded and `pod install`ed again. And it looks like it duplicated stuff. I'll fix that.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
paulrolfe, line 33: I've added tests for all of these new functions.

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
```diff
@@ -65,3 +65,11 @@ extension NSNumberFormatter {
         return decimalFormatter
     }
 }
+
+public extension Int {
```
brightredchilli: Not really keen on this being on all Ints...

NSTimeInterval + NSDateFormatter provides a good built in solution for this formatting imho

brightredchilli:
everything looks good to me except for the Int extension.

SwiftWisdom/Core/StandardLibrary/Numbers/Int+Extensions.swift
paulrolfe: That's a good point about using NSTimeInterval + NSDateFormatter. I'll just remove that function and corresponding test.

PR #50, opened 2016-09-13
ip_updateFooter - Allows replacing a footerview's content after creation

alexpersian:
- Adds a method for dynamically updating a UITableViews footer after it has been initially created.
- This comes in handy when using a custom view within a table's footer that needs to change after creation.

### DIFF
--- a/SwiftWisdom/Core/UIKit/UITableView+Extensions.swift
+++ b/SwiftWisdom/Core/UIKit/UITableView+Extensions.swift
@@ -1,0 +1,26 @@
+//
+//  UITableView+Extensions.swift
+//  SwiftWisdom
+//
+//  Created by Alexander Persian on 9/8/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import UIKit
+
+extension UITableView {
+    /**
+     Allows you to dynamically update a tableview's footer after it has initally been drawn. (Width is not needed since footers are always the full width of the tableview)
+
+     - parameter footerView: View that will be used in the footer
+     - parameter height:     Height of the footer within the tableview
+     */
+    public func ip_updateFooter(withView view: UIView, ofHeight height: CGFloat) {
+        self.tableFooterView = nil // Ensure that old views are not left over or stacked on top of each other
+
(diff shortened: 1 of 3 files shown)


SwiftWisdom/Core/UIKit/UITableView+Extensions.swift
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     Allows you to dynamically update a tableview's footer after it has initally been drawn. (Width is not needed since footers are always the full width of the tableview)
+
+     - parameter footerView: View that will be used in the footer
+     - parameter height:     Height of the footer within the tableview
+     */
+    public func ip_updateFooter(withView view: UIView, ofHeight height: CGFloat) {
+        self.tableFooterView = nil // Ensure that old views are not left over or stacked on top of each other
+
+        let footerContainer = UIView(frame: CGRect(x: 0, y: 0, width: self.ip_width, height: height)) // Footers are always the width of the table view
```
paulrolfe, line 21: Just a suggestion: you could have `height` be optional (with default value of nil) and if it's nil just use `view.height` as the height here. I'm thinking someone may have already set the height on their footer view being passed in -- in which case it's kind of redundant.

paulrolfe:
I know we use this in Monet, but not sure it's useful for everyone. Someone from outside the project should probably give this another 👀 

👍

SwiftWisdom/Core/UIKit/UITableView+Extensions.swift
brightredchilli: Whats the purpose of the footercontainer? Since the inner view is being constrained to all edges, footer container doesn't do anything - let me know if i'm missing something.

We should also write tests for this.

alexpersian:
Waiting until the Swift 3 migration is complete before coming back to make changes to this.

paulrolfe:
@alexpersian What's the status here? I'm leaning towards closing this -- I'm not sure the use case merits being in swift wisdom.

alexpersian:
@paulrolfe I can concede that it's not a general enough use case to be included here. Close away.

PR #52, opened 2016-09-16
update spec to specify swift version.

paulrolfe:
To help xcode8 projects out, we should specify the swift version for our pod.


paulrolfe:
Whoops, we're actually using swift 2.2

PR #53, opened 2016-09-16
Add Support for  2.3

cprime:


brightredchilli:
:ship:

paulrolfe:
I think you should specify this is swift 2.3 in the pod spec:
`s.pod_target_xcconfig = { 'SWIFT_VERSION' => '2.3' }`

SwiftWisdom/Core/DirectoryManager/DirectoryManager.swift
```diff
@@ -11,7 +11,7 @@ import Foundation
 public final class DirectoryManager {
     
     enum Error : ErrorType {
-        case UnableToCreatePath(url: NSURL)
+        case UnableToCreatePath(url: NSURL?) // TODO: review, 2.3 changes
```
Benuuu: What does this TODO mean?  please clarify it further

Benuuu:
:shipit:

SwiftWisdom/Core/DirectoryManager/DirectoryManager.swift
Benuuu: @brightredchilli

cprime: In swift 2.2, there was never a case such that this error would occur but the url would not exist. However, in swift 2.3, there are more cases in which NSURL creation can fail resulting in an error. The existing UnableToCreatePath error was modified to capture this case rather than creating a new url-less error case.

SwiftWisdom/Core/DirectoryManager/DirectoryManager.swift
```diff
@@ -11,7 +11,10 @@ import Foundation
 public final class DirectoryManager {
     
     enum Error : ErrorType {
-        case UnableToCreatePath(url: NSURL?) // TODO: When upgrading to 2.3 Swift, this type changed from NSURL -> NSURL?. Review to see if this is an appropriate change.
+        // TODO: When upgrading to 2.3 Swift, this type changed from NSURL -> NSURL?.
```
paulrolfe, line 14: I don't understand why this is a TODO still.

brightredchilli: I don't have the bandwidth to make to do a full review on this...unless colden wants to pick this up. Are you saying that it is acceptable to make the change to NSURL?

paulrolfe: I'm saying the change was made, so why is it a TODO?

brightredchilli: We made this change in very much of a hurry on Friday as we needed it for a release. The change was going to go out as is, with very little other eyes on it. The root of the matter is that DirectoryManager can fail in many other ways other than just 'UnableToCreatePath'. I think probably the best thing to do is to create the proper error states for them, rather than grouping them under 'UnableToCreatePath'...hence this comment.

brightredchilli: In the end it didn't end up getting shipped - so we have a bit more breathing room now, but not enough to look into separate error states and to create the appropriate test case scenarios for them - so either someone picks up on this thread or we can ship with the TODO. Maybe the wording should be refined?

paulrolfe: Ah, then I agree with you it seems we should add an error case `UnableToCreateURL(directoryName: String)` and leave this one as it was.

EDIT: Actually, I think this url should be optional if the NSURL's are optional now.

SwiftWisdom/Core/DirectoryManager/DirectoryManager.swift
```diff
@@ -86,9 +89,12 @@ extension NSFileManager {
...
         let pathsArray = NSSearchPathForDirectoriesInDomains(.DocumentDirectory, .UserDomainMask, true)
         guard let pathString = pathsArray.first, let documentsDirectoryPath = NSURL(string: pathString) else { fatalError("Unable to create directory") }
-        let directoryPath = documentsDirectoryPath.URLByAppendingPathComponent(directoryName)
-        try createDirectoryIfNecessary(directoryPath)
-        return directoryPath
+        if let directoryPath = documentsDirectoryPath.URLByAppendingPathComponent(directoryName) {
+            try createDirectoryIfNecessary(directoryPath)
+            return directoryPath
+        } else {
+            throw DirectoryManager.Error.UnableToCreatePath(url: nil)
```
paulrolfe, line 96: Here's where we should use a different error.

SwiftWisdom/Core/DirectoryManager/DirectoryManager.swift
cprime: do you have an alternative?

paulrolfe: @brightredchilli Ok, well it doesn't bother me that much. I'm fine leaving the todo.

paulrolfe:
:shipit:

SwiftWisdom/Core/DirectoryManager/DirectoryManager.swift
paulrolfe: Nevermind, we'll do this later.

PR #57, opened 2016-09-26
Update Rx subspec to Swift 3 syntax

Benuuu:


brightredchilli:
can you rebase against swift-3?

Benuuu:
I didn't actually test to see if the test worked because Tests aren't yet building. (it should, i hope)

vinnyt (approved):
:shipit:

SwiftWisdomTests/Rx/OperatorTests.swift
```diff
@@ -8,16 +8,17 @@
...
 
 class OperatorTests: XCTestCase {
 
     func testAddCompositeDisposable() {
         let compositeDisposable = CompositeDisposable()
-        let observable = Observable<String>.never()
+        let observable = Observable<String?>.never()
         XCTAssertEqual(0, compositeDisposable.count)
-        UILabel().rx_text <- observable >>> compositeDisposable
+        let _ = UILabel().rx.text <- observable >>> compositeDisposable
```
brightredchilli, line 13: question here - does this mean that everytime we use >>> we would need to do let _ ? that doesn't seem right?

Benuuu: @brightredchilli for composite disposables yes.  They changed the API for CompositeDisposable to return the key that references the disposable incase you want to dispose of it earlier.

brightredchilli: do we have a use case for this? wouldn't this cause a lot of unused variable warnings in our code? since the purpose of >>> is to compress our code, I think that it would be better not to return disposables here.

Intrepid.podspec
```diff
@@ -23,7 +23,7 @@ Pod::Spec.new do |s|
 
   s.subspec "Rx" do |rx|
     rx.source_files = "SwiftWisdom/Rx/**/**/*.swift"
-    rx.dependency 'RxSwift', '~> 2.6'
-    rx.dependency 'RxCocoa', '~>2.6'
+    rx.dependency 'RxSwift', '~> 3.0.0-beta.1'
```
vinnyt: beta.2 now.

vinnyt:
update to the latest rxswift and :shipit:

vinnyt (approved):

SwiftWisdomTests/Rx/OperatorTests.swift
Benuuu: The use case of a CompositeDisposable is slightly different than a DisposeBag.  The former you manage the lifecycle of manually and tell it to dispose when you are done with it.  The latter you typically just put a bunch of disposables into and forget about them.

Why would you want to remove a disposable from a CompositeDisposable?  I'm not entirely sure of the use case there, but the API seems to support it.  Maybe you'll have a subscription that would change as a result of something and you don't want it to be grouped anymore.

Do you think it would be better to drop this functionality for compression?

brightredchilli: I would think that the use case is, dispose upon an event(like view did appear/disappear), but also dispose together with the composite at some other point(deinit)

brightredchilli (approved):

brightredchilli:
I would vote to not remove the disposable but this shouldn't block the swift 3 module.

Benuuu:
Talked with @vinnyt about this.  We're going to favor dropping the DisposeKey return here.  FYI @brightredchilli

brightredchilli:
@Benuuu @vinnyt rebased onto swift-3, updated to beta2, ran tests, renamed OperatorTests to Rx+ExtensionsTest.

PR #64, opened 2016-09-30
Fix issue in Downloader where failed HTTP request still resulted in Success

mayasaxena:
- Added a check to `URLSession(session: NSURLSession, downloadTask: NSURLSessionDownloadTask, didFinishDownloadingToURL location: NSURL)` to ensure that the updater state only gets success if the HTTP response was also success. Otherwise, send an `HTTPRequestFailed` error to the updater.
- This provides an error to handle if the URL is invalid or if there is a server error


SwiftWisdom/Core/Downloader/Downloader.swift
```diff
@@ -82,3 +91,9 @@ extension NSURLSessionTask {
         return min(percentage, 1)
     }
 }
+
+extension NSHTTPURLResponse {
+    public var isSuccess: Bool {
+        return 200 ... 299 ~= statusCode
```
brightredchilli: wow. i really learned something new here.

Also, tiny note, remove the spaces between the numbers and ...

SwiftWisdom/Core/Downloader/Downloader.swift
```diff
@@ -9,57 +9,66 @@
...
-    
+
     public enum State {
         case Began
         case InProgress(CGFloat)
         case Completed(Result<NSURL>)
     }
-    
+
+    public enum Error : ErrorType {
```
brightredchilli: Just a heads up here - ErrorType is renamed Error in Swift 3. I don't know how we intend to handle this, but for now you can side step by renaming this DownloaderError. I know, super ugly :(

brightredchilli (approved):
Minor changes, then good to :ship: Please wait for one more review.

Also, I have no issues with this being merged into swift-wisdom on Swift 2.3(master), but please also apply your changes against the swift-3 branch to update for swift-3. Otherwise, we will not merge cleanly when swift-3 merges back to master. If you have questions on this, feel free to DM me!

brightredchilli:
Also, nice work.

Benuuu (approved):

PR #65, opened 2016-09-30
Update String

alanscarpa:


Podfile.lock
```diff
@@ -16,4 +16,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 1985213b54ee7ae4e9f82fc14acc408fd9e87934
 
-COCOAPODS: 1.1.0.rc.2
+COCOAPODS: 1.0.1
```
Benuuu: Please revert this change

alanscarpa: Done and updated.  But I don't like that i can still see this comment.

SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -1002,7 +998,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 8307DB003210EC8D96EBA9AF /* Pods-SwiftWisdom.debug.xcconfig */;
 			buildSettings = {
-				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = "$(inherited)";
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
Benuuu: Did you intentionally changes these or are they an artifact from 1.0.1 install?

alanscarpa: Artifact I assume.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -0,0 +1,48 @@
...
+        return String(substringCharacters)
+    }
+
+    public subscript(range: CountableClosedRange<Int>) -> String {
+        let chars = Array(characters)
+        let substringCharacters = chars[range]
+        return String(substringCharacters)
+    }
+    
+    public mutating func dropFirst() {
```
paulrolfe, line 26: Isn't is safer to use a prefix here?

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -0,0 +1,48 @@
...
+            let from16 = utf16.index(utf16.startIndex, offsetBy: nsRange.location, limitedBy: utf16.endIndex),
+            let to16 = utf16.index(from16, offsetBy: nsRange.length, limitedBy: utf16.endIndex),
+            let from = String.Index(from16, within: self),
+            let to = String.Index(to16, within: self)
+            else { return nil }
+        return from ..< to
+    }
+
+    // TODO: consider naming 'unchecked'
+    func ip_range(from range: ClosedRange<Int>) -> Range<String.Index> {
```
paulrolfe, line 43: Might also want one for  `range: CountableClosedRange<Int>`... I saw that issue a lot in doing NSData subscripts.

PR #66, opened 2016-09-30
Converts array and dictionary extensions

alexpersian:
- Converts Array extensions and Dictionary extensions
- Does not include test conversions


SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -122,11 +123,11 @@ extension Sequence where Iterator.Element: Equatable {
         return buffer
     }
     
-    public func ip_countOf(_ element: Iterator.Element) -> Int {
+    public func ip_countOf(element: Iterator.Element) -> Int {
         return self.filter { $0 == element } .count
     }
     
-    public func ip_containsAll<T: Sequence where T.Iterator.Element == Iterator.Element>(all: T) -> Bool {
+    public func ip_containsAll<T: Sequence>(all: T) -> Bool where T.Iterator.Element == Iterator.Element {
```
paulrolfe: Maybe this should just read like `ip_contains(all: ...)`... just trying to say remove the all from the func name since it will be in the first param. Alternatively hide the external parameter name so it's `ip_containsAll(...)` I'm not sure which our style favors.

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        XCTAssert(testArray.ip_containsAll([1, 12]))
+        XCTAssertFalse(testArray.ip_containsAll([14, 16]))
+        XCTAssertFalse(testArray.ip_containsAll([1, 5, 9, 10]))
+    }
+
+
+
+
+
+
```
paulrolfe, line 57: Extra lines

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
alexpersian: Thanks, missed this fix.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 public extension Array {
-    public func ip_subArrayFromIndices(_ indices: [Int]) -> [Element] {
+    public func ip_subArrayFrom(indices: [Int]) -> [Element] {
```
tomalley104: include from in the parameter `ip_subArray(fromIndices indices: [Int]) -> [Element]`

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -45,8 +45,8 @@ public extension Array where Element: Equatable {
      */
     public func ip_indicesOf(elements: [Element]) -> [Int] {
```
tomalley104: include "of" in the parameter:  `ip_indices(ofElements: [Element])`

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -122,11 +123,11 @@ extension Sequence where Iterator.Element: Equatable {
         return buffer
     }
     
-    public func ip_countOf(_ element: Iterator.Element) -> Int {
+    public func ip_countOf(element: Iterator.Element) -> Int {
```
tomalley104: include "of" in the parameter: `ip_count(ofElement: Iterator.Element)`

SwiftWisdom/Core/StandardLibrary/Array/Sequence+Utilities.swift
```diff
@@ -13,7 +13,7 @@ extension Sequence {
 }
 
 extension Sequence {
-    public func ip_splitFilter(filter: @noescape (Iterator.Element) throws -> Bool) rethrows -> (passed: [Iterator.Element], failed: [Iterator.Element]) {
+    public func ip_splitWith(filter: (Iterator.Element) throws -> Bool) rethrows -> (passed: [Iterator.Element], failed: [Iterator.Element]) {
```
tomalley104: include "with" in the parameter `ip_split(withFilter filter: (Iterator.Element) throws -> Bool)`

SwiftWisdom/Core/StandardLibrary/Dictionary/Dictionary+KeyPaths.swift
```diff
@@ -22,14 +22,14 @@ public extension Dictionary {
         
     }
     
-    public func ip_valueForKeyPath<T>(_ keyPath: String) -> T? {
-        var keys = keyPath.ip_keypathComponents()
+    public func ip_valueFor<T>(keypath: String) -> T? {
```
tomalley104: include 'for' in the parameter: `ip_value<T>(forKeypath keypath: String)`

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+
+import XCTest
+import SwiftiWisdom
+
+class ArrayUtilities: XCTestCase {
+
+    func testSubArrayFromIndicesOf() {
+        let testArray = [4, 8, 15, 16, 23, 42]
+        let indices = testArray.ip_indicesOf(elements: [15, 16, 19])
```
tomalley104, line 16: This test is dependent on `ip_indicesOf`; it could fail if `ip_indicesOf` doesn't do what you intend. So, I think you should hand-code the indices you're looking for.  

You can just do `let indices = [2, 3]` or even `.. = [2, 3, 20]` if you wanna still wanna see if you can trick it like you kinda are with that stray `19` in `ip_indicesOf`.

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        XCTAssert(testArray = [1, 5, 14, 12])
+    }
+
+    func testRemoveElements() {
+        var testArray = [1, 5, 9, 14, 12]
+        testArray.ip_remove(elements: [1, 5, 9])
+        XCTAssert(testArray == [14, 12])
+        testArray.ip_remove(elements: [12])
+        XCTAssert(testArray == [14])
+    }
```
tomalley104, line 38: maybe include a case where you try to remove elements that are not present in testArray like you are with `testRemoveObject()` above?

tomalley104: Also if you end up agreeing with the comment above to break `testRemoveObject` into multiple tests, do the same here.

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    func testRemoveObject() {
+        var testArray = [1, 5, 9, 14, 12]
+        var result = testArray.ip_remove(object: 9)
+        XCTAssertTrue(result)
+        XCTAssert(testArray = [1, 5, 14, 12])
+
+        result = testArray.ip_remove(object: 42)
+        XCTAssertFalse(result)
+        XCTAssert(testArray = [1, 5, 14, 12])
```
tomalley104, line 29: consider breaking this up into `testRemoveObject()` and `testRemoveNonExistantObject()` since you are testing two things here

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
alexpersian: I'm not sure if this is needed since `indices` is visible by default now. So it will read `ip_subArrayFrom(indices: [Int])`

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
alexpersian: Good point.

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
alexpersian: Good catch, thank you.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
tomalley104: I'm leaving these comments under the impression that swift 3 method naming convention is 'verb(PrepositionAndNouns)', i may be wrong.

tzm41:
Remove `SwiftWisdomTests/StandardLibrary/.DS_Store` file.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
alexpersian: After looking at the guidelines again I agree with you now. I do think it should be `verb(prepositionAndNouns noun)` like you wrote in your other comments. This makes the method body clearer.

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -22,55 +28,74 @@ public extension Array {
...
+    ///
+    /// - returns: Set of the array's elements
     public func ip_toSet() -> Set<Element> {
         return Set(self)
     }
 }
 
-extension Array {
+public extension Array {
+    //TODO: Evaluate switching from ip_safe to safe here
```
brightredchilli: I think we can remove this todo-since its unlikely we are going to not namespace extensions.

SwiftWisdomTests/StandardLibrary/DictionaryTests.swift
```diff
@@ -0,0 +1,25 @@
...
+class DictionaryTests: XCTestCase {
+
+    func testSetValueForKeypath() {
+        var testDictionary: Dictionary<String, Int> = [:]
+        testDictionary.ip_set(value: 42 as AnyObject, forKeypath: "hello")
+        XCTAssertEqual(testDictionary["hello"], 42)
+    }
+
+    func testValueForKeypath() {
+        let testDictionary: Dictionary<String, Int> = ["hello" : 42]
```
brightredchilli: for keypath test, it would be good to test nested dictionaries(since that is what the keypath is for), so have a dictionary of nested dictionaries - dont need to declare the type since swift will do that for you

```
let testDictionary = ["hello" : ["world : 42]]
testDictionary.ip_value(forKeypath: "hello.world")
```

brightredchilli (changes requested):
Will you make the test changes that @tomalley104 pointed out?

SwiftWisdomTests/StandardLibrary/DictionaryTests.swift
alexpersian: Sounds good, I'll change this. Guess I didn't realize the full functionality, haha.

brightredchilli:
bt dubs. really heroic effort here - please push your changes and rebase one more time onto nyc/standard-library so i can give a ship it.

alexpersian:
Thanks! Was a great way to get more familiar with the framework too. Thanks for leading the charge on starting this too!

SwiftWisdomTests/StandardLibrary/ArrayUtilitiesTests.swift
alexpersian: Done 👍

alexpersian:
Found some clearer documentation around extensions and how they work across modules, so I cleaned up the access levels here to reflect that.

brightredchilli (approved):
:ship:

PR #69, opened 2016-09-30
Update range function to return correct end index

mayasaxena:
- Updated range function to return correct `Range` from given `ClosedRange`
- This is causing a `ColorDescriptor` test (`testHex`) to fail


SwiftWisdom/Core/StandardLibrary/String/String+NSRange.swift
```diff
@@ -23,9 +23,10 @@ extension String {
 
     // TODO: Need to write tests for this
     // TODO: consider naming 'unchecked'
+    // TODO: Ying
```
brightredchilli: can we have this say "ying to write tests and resolve any merge conflicts"...lol this will give me a heart attack some time in the future when i forget what this is about :D

brightredchilli (approved):

SwiftWisdom/Core/StandardLibrary/String/String+NSRange.swift
```diff
@@ -23,9 +23,10 @@ extension String {
 
     // TODO: Need to write tests for this
```
Benuuu: Remove this TODO: since we have the other Ying to write tests

Benuuu (approved):

brightredchilli:
this has two ship-its @mayasaxena I merged this into swift-3. Thanks!

PR #74, opened 2016-10-05
updated NSAttributedStrings

brightredchilli:


SwiftWisdom/Core/StandardLibrary/NSAttributedString/NSMutableAttributedString+Format.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-extension NSMutableAttributedString {
+public extension NSMutableAttributedString {
```
alexpersian: I was wrong about this when we discussed it last week. After digging some more I found that extensions will inherit the access level of their type, but methods inside can set their own access level. So long as the functions we want accessible from outside are declared `public` we don't need this specifier. (based on [this link](https://www.andrewcbancroft.com/2015/04/22/3-nuances-of-swift-extensions/#extension-visibility) and [apple docs](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID25))

I'll update my PR to reflect this as well.

SwiftWisdom/Core/StandardLibrary/NSAttributedString/NSMutableAttributedString+Format.swift
```diff
@@ -48,16 +47,39 @@ extension NSMutableAttributedString {
...
-            self.replaceCharacters(in: insertRange, with: arg)
+            replaceCharacters(in: insertRange, with: arg)
             rangeLimit = NSMakeRange(0, insertRange.location)
             insertRange = str.range(of: "%@", options: .backwards, range: rangeLimit)
         }
     }
 }
+
+// TODO: move to String extensions.
+fileprivate extension String {
```
alexpersian, line 81: Does this need to be moved if it's only being used in this file?

SwiftWisdomTests/StandardLibrary/NSAttributedString/NSMutableAttributedString+FormatTests.swift
```diff
@@ -0,0 +1,61 @@
...
+    private func assert<T: Equatable>(attributedString: NSAttributedString,
+                        atIndex index: Int,
+                        attributeName: String,
+                        equals expected: T) {
+        let attributes = attributedString.attributes(at: index, effectiveRange: nil)
+        let actual = attributes[attributeName] as! T
+        XCTAssert(actual == expected)
+    }
+
+    func testExample() {
```
alexpersian: Can this be removed since it's just an example?

alexpersian (changes requested):
Big mixup on my part in regards to the `public` extension stuff. Sorry about that! Other than that just saw a couple things.

SwiftWisdom/Core/StandardLibrary/NSAttributedString/NSMutableAttributedString+Format.swift
brightredchilli: the doc says "If you extend a public or internal type, any new type members you add have a default access level of internal." - so I'm not actually so sure on this note. Let's not worry about this for now, just merge in your changes, we will test before a merge.

SwiftWisdomTests/StandardLibrary/NSAttributedString/NSMutableAttributedString+FormatTests.swift
brightredchilli: yes absolutely

SwiftWisdom/Core/StandardLibrary/NSAttributedString/NSMutableAttributedString+Format.swift
brightredchilli: I would like to make this public - but dont want to conflict with the work alan is doing. will take care of this before nyc/standard-library gets merged,  updating comment to do so

alexpersian (approved):

SwiftWisdom/Core/StandardLibrary/NSAttributedString/NSMutableAttributedString+Format.swift
```diff
@@ -48,16 +47,39 @@ extension NSMutableAttributedString {
...
+    // See ip_format(arguments: [NSAttributedString])
+    func ip_format(withArguments arguments: NSAttributedString...) {
+        ip_format(withArguments: arguments)
+    }
+
+    /**
+     Assumes that the received is a attributed string with format placeholders "%@". Inserts the
+     NSAttributedString in args into each of the placeholders.
+
+     - parameter args: comma separated list of arguments (NSAttributedString)
```
Benuuu: This line appears to be outdated.

Benuuu (approved):

SwiftWisdom/Core/StandardLibrary/NSAttributedString/NSMutableAttributedString+Format.swift
brightredchilli: just had to rename the comment, fixed.

PR #80, opened 2016-10-24
Alex/further cleanup

alexpersian:
- Moves the custom operators out of the global scope now that Swift 3 supports overloaded operators within extensions.
- Updates the project settings to comply with Xcode 8's recommendation.


vinnyt:
are there tests that need to be updated for these. If not :shipit:

alexpersian:
Outside of the pre-existing tests, I didn't add any myself in this PR.

I plan to add more test coverage in a future PR.

SwiftWisdom/Core/Version/Version.swift
```diff
@@ -33,45 +33,47 @@ extension Version : CustomStringConvertible {
...
 
-public func ==(lhs: Version, rhs: Version) -> Bool {
-    guard lhs.major == rhs.major else { return false }
-    guard lhs.minor == rhs.minor else { return false }
-    guard lhs.patch == rhs.patch else { return false }
-    return true
-}
+extension Version : Equatable {
+    public static func ==(lhs: Version, rhs: Version) -> Bool {
+        guard lhs.major == rhs.major else { return false }
```
chow-land, line 40: couldn't you just ::

```
 return
    lhs.major == rhs.major
    && lhs.minor == rhs.minor
    && lhs.patch == rhs.patch
```

SwiftWisdom/Core/Downloader/Downloader.swift
```diff
@@ -82,3 +74,15 @@ extension URLSessionTask {
         return min(percentage, 1)
     }
 }
+
+// MARK: Operator Overloads
+
+extension Downloader {
+    public static func ==(lhs: Downloader, rhs: Downloader) -> Bool {
```
Benuuu: Will you update these extensions (and the others in this PR) to include a space after the operator?

SwiftWisdom/Core/Version/Version.swift
```diff
@@ -33,45 +33,47 @@ extension Version : CustomStringConvertible {
...
+        guard lhs.minor == rhs.minor else { return false }
+        guard lhs.patch == rhs.patch else { return false }
+        return true
+    }
 
-public func !=(lhs: Version, rhs: Version) -> Bool {
-    return !(lhs == rhs)
+    public static func !=(lhs: Version, rhs: Version) -> Bool {
+        return !(lhs == rhs)
```
chow-land, line 47: I'd opt for "return lhs != rhs" since I think it's a little more clear, but to each their own.

Benuuu (approved):

SwiftWisdom/Core/Downloader/Downloader.swift
alexpersian: Sure

SwiftWisdom/Core/Version/Version.swift
```diff
@@ -33,45 +33,47 @@ extension Version : CustomStringConvertible {
...
 
-public func > (lhs: Version, rhs: Version) -> Bool {
-    if lhs.major != rhs.major {
-        return lhs.major > rhs.major
-    } else if lhs.minor != rhs.minor {
-        return lhs.minor > rhs.minor
-    } else {
-        return lhs.patch > rhs.patch
+    public static func > (lhs: Version, rhs: Version) -> Bool {
+        if lhs.major != rhs.major {
```
chow-land, line 71: can this not be reduced to:
`return lhs.major > rhs.major  || lhs.minor > rhs.minor || lhs.patch > rhs.patch`

bobgilmore: No, that suggested change would make `2.2.5 > 3.3.3` return `true` when it should be `false`

chow-land: ah, I didn't realize that Version actually represented an "N.N.N" format number, good call.

jfyles: Just out of curiosity, why not convert to a string and string compare with NSStringCompareOptions.NumericSearch? I feel like that would be flexible if the format changed for some reason in the future (you'd only need to change the tostring, not the comparison operator overloads)

bobgilmore:
Once the spacing issues in function declarations (i.e., replacing `==(` with `== (` ) are addressed, I'm good with the changes.  There are some good ideas in here for some _re_-implementations of the algorithms, but those shouldn't happen in this PR - let's keep this one scoped to simply "moving" the implementations into the "correct" place.

alexpersian:
Noting down the conversation and proposed changes for a separate PR. Tracked with #81

PR #84, opened 2016-11-08
Adds a utility function to generate a transformation between two sets

cprime:


bobgilmore:
Is there a major advantage to doing this rather than creating `Sets`, and then using either `subtracting` or ( `symmetricDifference` , followed by `subtracting` ) to get the exact difference you want?

bobgilmore:
OK, to answer my own question: the implementation in this PR requires that `T` conform to `Equatable`, but using the `Set` mechanism that I suggested above would _further_ require that `T` conform to `Hashable`. That's not ideal.

vinnyt:
It feels like this should be an extension as opposed to a method on the array class.  I also feel like the name isn't exactly clear.   What is the reasoning behind the static method vs. extension?

I would expect it to behave something like the following:
`let difftuple = startArray.changes(to: newArray)`

cprime:
@vinnyt I did not have great reasons. I've switched to a member function which simplified both the signature and the callsite.

paulrolfe:
I'm probably complicating this more than needed. But when I think about comparing changes, I usually start with the new one and compare it to the old one. So the ideal usage to me would be the opposite of how it is currently.
`let changes = newArray.changes(from: oldArray)`
Maybe we could have both? @cprime

brightredchilli:
I feel we should just have one method instead of a `new.from(old)` and `old.to(new)` methods. I like 'new.changes(from: old)` though, but not a blocker.

brightredchilli (approved):

bobgilmore:
Ah, now I remember why I started thinking about `Set`s in the first place.

I've used (and written) methods like this in the past, and invariably me (and others) think that they do more than than advertised.  In particular, clients start to think that `old + added = new`, but that's _not_ the case, because of ordering and duplicates.  Handling duplicates is especially tricky, since different implementations return different results.  

Of course, those are _exactly_ the issues that thinking in terms of `Set`s resolves, because `Set`s are unordered and have no duplicate entries. So the results of `subtracting` and `symmetricDifference` are unambiguous (or at least, can't be misleading). I suspect that's why the library offers the `Set` methods, and no corresponding `Array` methods, in the first place.

The documentation is correct ("return the list of added elements and the list of removed elements") but given the ambiguities, I don't think this is a good candidate for a "generic" reusable implementation.  If someone needs an array difference, maybe they should have to think about the issues and implement something specific to their application?

paulrolfe:
:shipit: Can you also bump the spec version to 0.6.2? People have been missing that so I can't get any of the new updates in my project... 👿

brightredchilli:
@bobgilmore i feel like youre making some very important points that I haven't fully grasped. Is the concern(for example) that substraction when there are two duplicate entries on an array is ambiguous? (It can be that both entries are substracted, or only the first one, or the last one)? Is this what you mean by 'different implementations return different results'?

@cprime is there a reason you can't use Set for your purposes?

cprime:
@brightredchilli @bobgilmore only that it would require the elements conform to Hashable. If we were to go that route, then it probably would make sense to shift it completely to a Set extension, which is more accurate semantically.

bobgilmore:
@brightredchilli Yes, that thing that you mentioned regarding duplicate entries is one example of the kind of thing I'm concerned about.

@cprime If you WERE to shift it to a Set extension, corresponding methods ALREADY exist in the Swift Standard Library.  Your `removed` would "turn into" the already-existing `subtracting` method, and `added` would be like doing a `union` and then a `subtracting`

cprime:
@brightredchilli @bobgilmore I have updated the logic to use sets instead, but now that this change has been made, the function is probably trivial enough to just close the PR...

bobgilmore:
Yeah, now that we've reached this point, I'd be inclined to just kill the PR.  😢  

But @paulrolfe correctly observed that we should be updating the spec version more often, let's make sure that we take that piece of advice going forward!

paulrolfe:
I'm fine with closing this too.

PR #86, opened 2016-11-11
Data extensions

paulrolfe:
- Created Data extensions to mirror NSData extensions
  - NSData remains in place -- some names were adjusted to indicate that it deals with `NSData` as opposed to `Data`
  - Not all functionality needed to be duplicated because `Data` has much of that built in. Like prefixing, suffixing, and trimming.
  - Includes tests for all Data extensions
- Moved some folders around...
  - make Foundation include `NSAttributedString`
  - Consolidates some String_Data extensions in one place (within Strings folder)
- Using `inherited` flag to shut up the pod install warnings.


SwiftWisdom/Core/Foundation/Data+Extensions.swift
```diff
@@ -0,0 +1,138 @@
...
+//
+//  Created by Paul Rolfe on 11/10/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+
+// MARK: Conversion
+
+public extension Data {
```
alexpersian: I think we should leave extensions on default types defined as `extension Type` without a specific access modifier unless necessary. In this case the `public` modifier is redundant because of the `public` declaration of the functions already expose them to users of the framework.

Reference: [Apple Extensions Doc](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/AccessControl.html#//apple_ref/doc/uid/TP40014097-CH41-ID25)

alexpersian (changes requested):
Looks good minus one comment.

alexpersian (approved):

vinnyt:
:shipit:

PR #87, opened 2016-11-15
Update Downloader to complete with failure if HTTP request was not success

mayasaxena:
- Added a check to `URLSession(session: NSURLSession, downloadTask: NSURLSessionDownloadTask, didFinishDownloadingToURL location: NSURL)` to ensure that the updater state only gets success if the HTTP response was also success. Otherwise, send an `HTTPRequestFailed` error to the updater.
- This provides an error to handle if the URL is invalid or if there is a server error 
- Swift 3 version of earlier PR

### DIFF
--- a/SwiftWisdom/Core/Downloader/Downloader.swift
+++ b/SwiftWisdom/Core/Downloader/Downloader.swift
@@ -15,6 +15,10 @@ public final class Downloader : NSObject, URLSessionDownloadDelegate {
         case inProgress(CGFloat)
         case completed(Result<URL>)
     }
+
+    public enum DownloaderError : Error {
+        case httpRequestFailed
+    }
     
     public let url: URL
     public let id = UUID().uuidString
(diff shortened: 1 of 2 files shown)


paulrolfe (approved):

paulrolfe:
Be sure to update the spec version to `0.6.3` and tag the commit as such.

SwiftWisdom/Core/Downloader/Downloader.swift
```diff
@@ -52,6 +56,11 @@ public final class Downloader : NSObject, URLSessionDownloadDelegate {
     // MARK: NSURLSessionDownloadDelegate
     
     public func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
+        if let httpResponse = downloadTask.response as? HTTPURLResponse, !httpResponse.isSuccess {
```
brightredchilli, line 59: does updater not need to get set to nil in the failure case?

SwiftWisdom/Core/Downloader/Downloader.swift
```diff
@@ -52,6 +56,12 @@ public final class Downloader : NSObject, URLSessionDownloadDelegate {
...
+            updater?(.completed(.failure(DownloaderError.httpRequestFailed)))
+            updater = nil
```
brightredchilli: @mayasaxena Thanks for the update, but I think that this line seems repeated with the one below now, and the early return obscures this fact. Can you leave the updater = nil assignment to the end, and put success in the else-block?

brightredchilli (approved):

PR #94, opened 2016-11-20
Changes string+indexing methods to be public

alexpersian:
* Updates access levels on `String+Indexing.swift`
* Adds missing tests for `ip_dropFirst`


SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -1,7 +1,6 @@
 import Foundation
 
-public extension String {
-    // TODO: Need to write tests for this
```
brightredchilli, line 4: ~~Can you make this TODO more specific instead of removing it? In particular, `ip_range(from range: NSRange) -> Range<String.Index>?` remains untested, as far as i can see. The rest have tests~~ Never mind, I read it wrong, all good.

brightredchilli (approved):
Thanks for patching this up!

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
alexpersian: Great!

paulrolfe:
I just need to point out that this missed a pod spec version bump.

alexpersian:
Crap, that was my bad.

PR #95, opened 2016-11-21
Bumps the podspec

alexpersian:


Benuuu (approved):

brightredchilli:
Worth mentioning here - this bump does not do anything unless we also update the cocoapods trunk. Our cocoapod is still at 0.6.3.

paulrolfe:
@brightredchilli That's right and it also requires adding the version tag to github. So I tagged this and pushed to trunk for `0.6.4`.

Issue #90, opened 2016-11-17
Add test coverage for String+Indexing extension


andrewdolce:
From TODO comment in String+Indexing.swift. Make sure all public API functions are covered by a test.

alexpersian:
Closing this issue as fixed through #94

PR #100, opened 2016-11-28
Made sure of proper test coverage for String+Indexing extension

Dennis-Cherchenko:


SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
```diff
@@ -2,6 +2,30 @@ import XCTest
 
 class String_IndexingTests: XCTestCase {
     
+    func testRangeSubscriptReturnsProperString() {
+        XCTAssertEqual("hello"[0..<0], "")
```
paulrolfe, line 6: I think we should also test strings with emojis, just to be robust.

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
```diff
@@ -18,34 +42,20 @@ class String_IndexingTests: XCTestCase {
...
-        XCTAssertEqual("ello", testString)
-    }
-
-    func testIpDropFirstWithEmptyString() {
-        var emptyString = ""
-        emptyString.ip_dropFirst()
-        XCTAssert(emptyString.isEmpty)
+    func testIpFullRange() {
+        var testString = ""
+        assertIpRange(onString: testString, withNSRange: testString.ip_fullrange)
```
bobgilmore: Shouldn't `testIpFullRange` simply verify that the correct NSRange is generated on a call to `testString.ip_fullrange`, rather than *also* test the functionality of `ip_range` via the `assertIpRange` call?

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
```diff
@@ -2,6 +2,30 @@ import XCTest
 
 class String_IndexingTests: XCTestCase {
     
+    func testRangeSubscriptReturnsProperString() {
+        XCTAssertEqual("hello"[0..<0], "")
+        XCTAssertEqual("hello"[0..<1], "h")
+        XCTAssertEqual("hello"[0..<5], "hello")
```
bobgilmore, line 8: Do we also need some kind of out-of-bounds test (i.e., `XCTAssertSOMETHING("hello"[0..<6], "hello")`) to ensure that the desired behavior (whatever that is - return the full string, throw a specific error, probably NOT seg-fault) doesn't break in the future?

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
bobgilmore: If we're doing something funky w.r.t. string length and indexing, sure.  Otherwise, aren't we just adding a test that Swift's string indexing works with multibyte characters?

Also I assume that *any* high-byte Unicode character would do for testing?

alexpersian:
Curious what the purpose of this PR is since #94 addressed the test coverage and removed the `// TODO:`. I probably should have closed that issue #90

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
paulrolfe: Looks like it's just using `characters` property on string, no special indexing by bytes, so it's probably fine as is.

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
andrewdolce: @bobgilmore Dennis and I talked about this a little. It seems like this function was originally written so that an out-of-bounds access would crash at runtime, much like array subscripting. I see the value in handling out-of-bounds in a better way, but didn't necessarily want to change the interface.

Some ideas are:
- Keep it as-is but add some alternative subscript(s) with safer behavior.
- Throw an error. This would break the existing interface.
- Change return type to `String?` and return `nil` when out-of-bounds. This also would break the existing interface.
- "Clamp" the given range to something in bounds and return the resultant string. This doesn't break the interface, but I'm not sure if that's the behavior most people would expect.

SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -1110,7 +1110,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 8307DB003210EC8D96EBA9AF /* Pods-SwiftWisdom.debug.xcconfig */;
 			buildSettings = {
-				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = "$(inherited)";
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
andrewdolce, line 1113: Any idea what introduced this change? If it's not necessary, we should revert.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -2,17 +2,31 @@ import Foundation
...
         let chars = Array(characters)
         let substringCharacters = chars[range]
         return String(substringCharacters)
     }
+    
+    public subscript(safe range: Range<Int>) -> String {
+        let chars = Array(characters)
+        let substringCharacters = chars[range]
+        return String(substringCharacters)
+    }
```
andrewdolce, line 26: Guessing this is work-in-progress? I'm OK with adding a "safe" access alternative but looks like the implementation here hasn't changed from the original.

SwiftWisdom.xcodeproj/project.pbxproj
Dennis-Cherchenko: I think this is cause by when you accept the changes from the warning notification Xcode give to "update project to recommended settings"

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
brightredchilli: I have seen the 'safe behaviour' implemented in swift internal apis with 'safe:' as the parameter name. Some of our own extensions use this, like Data extensions. I would vote for keeping it this way, with OOB crashing, and leave it up to the caller to manage.

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
brightredchilli: While not in this PR, I would love if we can take on @bobgilmore's concern and just try seeing if the indexing works on these two(bob, feel free to chime in if you have specific examples in mind)

1. la ingeniería (assert that [0...12] returns full string)
2. 工程师 (assert that [0...3] returns full string)

If those don't work, it's worth to document that as gotchas. @Dennis-Cherchenko if you don't end up doing this can you add a github issue?

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
bobgilmore: Yeah, you're right, crash WOULD be the right thing to do here.

paulrolfe:
As Dennis is gone, and as Alex pointed out, much of this work was done already, I'm going to close this.

PR #101, opened 2016-12-05
Fixes for Swift 3

Benuuu:
- Fixes RxSwift extensions to be more generic
- Fixes an issue when compiling with a mixed obj-c and swift codebase

### DIFF
--- a/SwiftWisdom/Rx/Rx+Extensions.swift
+++ b/SwiftWisdom/Rx/Rx+Extensions.swift
@@ -23,18 +23,20 @@ infix operator <-> : Binding
 infix operator <- : Binding
 infix operator >>> : Disposing
 
-public func <- <T>(property: AnyObserver<T>, variable: Observable<T>) -> Disposable {
-    return variable
-        .observeOn(MainScheduler.instance)
-        .bindTo(property)
+public func <- <T: ObserverType, O: ObservableType>(observer: T, observable: O) -> Disposable where T.E == O.E {
+    return observable.observeOn(MainScheduler.instance).bindTo(observer)
 }
 
-public func <- <T>(property: AnyObserver<T>, variable: Variable<T>) -> Disposable {
-    return property <- variable.asObservable()
+public func <- <T: ObserverType, O>(observer: T, variable: Variable<O>) -> Disposable where T.E == O {
+    return observer <- variable.asObservable()
 }
 
-public func <- <T>(variable: Variable<T>, property: ControlProperty<T>) -> Disposable {
-    return property.bindTo(variable)
+public func <- <T, O: ObservableType>(variable: Variable<T>, observable: O) -> Disposable where O.E == T {
+    return observable.bindTo(variable)
+}
+
+public func <- <T>(observer: Variable<T>, observable: Variable<T>) -> Disposable {
+    return observer <- observable.asObservable()
 }
 
 public func >>> (disposable: Disposable, disposeBag: DisposeBag) {
--- a/SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
+++ b/SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
@@ -6,6 +6,43 @@ import RxCocoa
 
 class OperatorTests: XCTestCase {
 
+    func testVariableBinding() {
+        let disposeBag = DisposeBag()
+        let label = UILabel()
+        let variable = Variable<String?>(nil)
+
+        label.rx.text <- variable >>> disposeBag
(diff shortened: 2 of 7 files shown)


SwiftWisdom/Rx/Rx+Extensions.swift
```diff
@@ -23,17 +23,15 @@ infix operator <-> : Binding
...
-public func <- <T>(property: AnyObserver<T>, variable: Observable<T>) -> Disposable {
-    return variable
-        .observeOn(MainScheduler.instance)
-        .bindTo(property)
+public func <- <T: ObserverType, O: ObservableType>(property: T, variable: O) -> Disposable where T.E == O.E {
+    return variable.observeOn(MainScheduler.instance).bindTo(property)
 }
 
-public func <- <T>(property: AnyObserver<T>, variable: Variable<T>) -> Disposable {
+public func <- <T: ObserverType, O>(property: T, variable: Variable<O>) -> Disposable where T.E == O {
```
paulrolfe: Do the tests need to be updated for this?

SwiftWisdom/Rx/Rx+Extensions.swift
```diff
@@ -23,17 +23,15 @@ infix operator <-> : Binding
 infix operator <- : Binding
 infix operator >>> : Disposing
 
-public func <- <T>(property: AnyObserver<T>, variable: Observable<T>) -> Disposable {
-    return variable
-        .observeOn(MainScheduler.instance)
-        .bindTo(property)
+public func <- <T: ObserverType, O: ObservableType>(property: T, variable: O) -> Disposable where T.E == O.E {
```
brightredchilli: aren't these considered operators as well? Do they not need nonobjc?

brightredchilli (commented):
Tests appear to fail..

SwiftWisdom/Rx/Rx+Extensions.swift
Benuuu: It could be because they're not in an extension on ObserverType.  I will look into moving them there now.

SwiftWisdom/Rx/Rx+Extensions.swift
Benuuu: There is actually only one test for composite disposable here.

The tests wouldn't need to be updated because the functions are made more generic.

I will add some tests for these.

Benuuu (commented):
Looking into the jenkins issue now.

SwiftWisdom/Rx/Rx+Extensions.swift
Benuuu: Ran into some issues with the compiler when trying to move these to extensions (not related to nonobjc).  I was able to test it out on `<->` and it resulted in the compiler complaining that "Only methods, initializers, properties and subscript declarations can be declared nonobjc"

It looks like this is necessary for Downloader and NSDate because they inherit from NSObject whereas the RxSwift classes do not, so we do not need to specify that you cannot access these from objc

paulrolfe (approved):
:shipit:

Benuuu:
retest this please

SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
```diff
@@ -6,6 +6,28 @@ import RxCocoa
 
 class OperatorTests: XCTestCase {
 
+    func testVariableBinding() {
+        let disposeBag = DisposeBag()
+        let label = UILabel()
+        let variable = Variable<String?>(nil)
+        XCTAssertEqual(nil, label.text)
```
brightredchilli: is this assert necessary? it is pre-interaction with subject, right?

SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
```diff
@@ -6,6 +6,28 @@ import RxCocoa
...
+
+        label.rx.text <- variable >>> disposeBag
+        XCTAssertEqual(nil, label.text)
+
+        variable.value = "hello"
+        XCTAssertEqual("hello", label.text)
```
brightredchilli, line 18: could we also add an assert here and below to check that setting variable.value to nil sets label.text to nil as well?

brightredchilli (approved):
2 quick comments

SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
Benuuu: I don't think it is necessary.  It is pre-interaction with subject that is asserting default behavior.  I will remove it.

PR #103, opened 2016-12-21
Add Array extension to separate an array's elements into chunks.

Tom-Le:
Also fix a unit test not compiling due to some number literals not fitting into a normal integer, and fix a minor spelling mistake in README.

### DIFF
--- a/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
+++ b/SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
@@ -32,6 +32,43 @@ extension Array {
     }
 }
 
+extension Array {
+    /// Separates this array's elements into chunks with a given length. Elements in these chunks and the chunks themselves
+    /// are ordered the same way they are ordered in this array.
+    ///
+    /// If this array is empty, we always return an empty array.
+    ///
+    /// If `length` is not a positive number or if it's greater than or equal to this array's length, we return a new array
+    /// with no chunks at all, plus the "remainder" which is this array itself iff `includeRemainder` is `true`.
+    ///
+    /// - parameter length: Number of elements in each chunk.
+    /// - parameter includeRemainder: If set to `true`, include the remainder in returned array even if its length is less than
+    ///   `length`. If set to `false`, discard the remainder. Defaults to `true`.
+    ///
+    /// - returns: New array of chunks of elements in this array.
+    public func ip_chunks(of length: Int, includeRemainder: Bool = true) -> [[Element]] {
+        guard !isEmpty else { return [] }
+        guard length > 0 else { return includeRemainder ? [self] : [] }
+
+        var chunks = [[Element]]()
+        var nextChunkLeftIndex = 0
+        let nextChunkRightIndex = { nextChunkLeftIndex + length - 1 }
+
+        while nextChunkRightIndex() < count {
+            let nextChunk = Array(self[nextChunkLeftIndex...nextChunkRightIndex()])
+            chunks.append(nextChunk)
+            nextChunkLeftIndex += length
+        }
+
+        if includeRemainder && nextChunkLeftIndex < count {
+            let remainder = Array(self[nextChunkLeftIndex..<count])
+            chunks.append(remainder)
+        }
+
+        return chunks
+    }
+}
+
 extension Array where Element: Equatable {
     /// Removes a single element from the array.
     ///
--- a/SwiftWisdomTests/StandardLibrary/Numbers/Integer+ExtensionsTests.swift
+++ b/SwiftWisdomTests/StandardLibrary/Numbers/Integer+ExtensionsTests.swift
@@ -26,7 +26,7 @@ class IntegerExtensionTests: XCTestCase {
     }
     
     func testToMagnitudeString() {      
-        let numbers = [14, 1234, 15420003, 5678000000, 8888000000000, 999999, 7666555444333222, 1234567891234567890]
+        let numbers: [Int64] = [14, 1234, 15420003, 5678000000, 8888000000000, 999999, 7666555444333222, 1234567891234567890]
         let formattedNumbersOnePlace = ["14", "1.2K", "15.4M", "5.7B", "8.9T", "1M", "7.7Q", "1.2E18"]
         let formattedNumbersTwoPlaces = ["14", "1.23K", "15.42M", "5.68B", "8.89T", "1M", "7.67Q", "1.23E18"]
         
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.6.5"
+  s.version       = "0.6.6"
   s.summary       = "Swift Bag"
   s.description   = <<-DESC
                     Collection of extensions and utility classes by and for the developers at Intrepid Pursuits.
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Build Status](http://ci.intrepid.io:8080/buildStatus/icon?job=Swift Wisdom)](http://ci.intrepid.io:8080/job/Swift%20Wisdom/)
 [![Coverage](http://ci.intrepid.io:9913/jenkins/cobertura/Swift%20Wisdom/)](http://ci.intrepid.io:8080/job/Swift%20Wisdom/cobertura/)
 
-This library is mean to be a general grab bag of Swift methods made by the engineers at IntrepidPursuits.  As this library goes, useful components will be isolated into separate podspecs where appropriate.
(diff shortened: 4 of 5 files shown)


bobgilmore:
Setting aside (for the moment) the issue of zero or negative `length` parameter; I'd prefer to remove the documented "special case" handling of `length` > array's length.  In this case, following the other rules, we would allocate *zero* chunks of length `length`, and add the "remainder" (the entire array), or not, according to the `includeRemainder` rule.

That is, if `length` > array's length and `includeRemainder` is `false`, return an *empty* array. OTOH, if `length` > array's length and `includeRemainder` is `true`, go ahead and return an array containing the original array as its lone element.

bobgilmore:
Another thing; this proposed implementation returns elements in the *same order* as they were in the original set.  For example, if you have

```
[1, 2, 3, 4, 5, 6, 7, 8].ip_chunks(ofLength: 3)
```

the current implementation returns

```
[[1, 2, 3], [4, 5, 6], [7, 8]] 
```

Since that's the case, then whether or not you intend it, your clients *will* come to rely on that ordering, and their code will break if it ever changes.  I've seen this happen with MATLAB functions that happened to sort their inputs as an implementation side-effect. When MathWorks switched to more efficient implementations that no longer sorted automatically, customers were incredibly displeased that the unstated sorting "contract" was broken.

I recommend either...
1. Documenting that the order of elements is preserved (I can help with wording for that), and *commit* to not changing it, or
2. Documenting that the order of the return values is *not guaranteed and subject to change at any time.*

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -32,6 +32,37 @@ extension Array {
...
+    public func ip_chunks(ofLength length: Int, includeRemainder: Bool = true) -> [[Element]] {
+        guard length > 0 else { return [self] }
+
+        var mutable = self
+        var chunks = [[Element]]()
+
+        while mutable.count >= length {
+            let next = Array(mutable.prefix(length))
+            chunks.append(next)
+            mutable.removeFirst(length)
```
bobgilmore: I'm a little concerned about the performance characteristics of this implementation.  (I wouldn't necessarily be concerned for a known use case in a particular product, but since we're proposing a general purpose library function, I want to ensure that it's performant.)

I assume that `removeFirst` requires an allocation and an array copy for each call?  If we expect to be using this for large arrays (or for array / `length` combinations that will hit this line many times), would we be better off to use another algorithm that peels elements off the *end* of the array?  Perhaps even better, avoid making `remove*` calls altogether and track the  the region to copy "by hand," avoiding the additional allocations and copies that `removeFirst` presumably introduces?

(BTW, this is exactly why I'm worried about ensuring order, or not, in my "overall" comment in this PR.  Because, for example, we *could* ship this implementation, which guarantees order, and then decide to move to a "reverse" implementation that reverses the order of the blocks that are peeled off.  It would be best to decide on, and *document,* a strategy once.)

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -32,6 +32,37 @@ extension Array {
...
+    ///
+    /// If `length` is not a positive number or if it's greater than or equal to this array's length, we return a new array
+    /// whose sole element is this array itself.
+    ///
+    /// - parameter length: Maximum number of elements in each chunk.
+    /// - parameter includeRemainder: If set to `true`, include the last chunk in returned array even if its length is less than
+    ///   `length`. If set to `false`, discard the last chunk if its length is less than `length`. Defaults to `true`.
+    ///
+    /// - returns: New array of chunks of elements in this array.
+    public func ip_chunks(ofLength length: Int, includeRemainder: Bool = true) -> [[Element]] {
```
vinnyt: I think a signature of `ip_chunks(of length: Int, includeRemainder: Bool = true) ...`

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
Tom-Le: You're right! In hindsight making `remove*` calls seems like a bad idea especially for very large arrays. I propose that we document that the order of the elements is preserved (though I'll definitely need help with wording for that!) and I'll change the implementation to track the region to copy by hand.

I also agree that we should handle the negative or zero `length` case like you described (add the "remainder"/the entire array if `includeRemainder` is true, return an empty array otherwise). About the `length > array.count` case though, since I documented that `length` is the *maximum* length of a chunk, should I always include the entire array in the returned result regardless of what `includeRemainder` is? Or should I document `length` as the precise length of each chunk except for the very last one (whose length can be less than or equal)?

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
Tom-Le: Would you mind elaborating a little further? I'm not sure if I follow…

paulrolfe: first external param `of`, internal would be `length`

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
bobgilmore: To address your question about the `length` > `array.count` case - I think that the correct way to think of it is to just "do the math."  

In the "normal" case, you're returning `array.count / length` chunks, and then one additional chunk of the "leftovers" if (`array.count % length > 0` AND `includeRemainder == true`).

I think that's exactly what you should do in the `length` > `array.count` case.  In that case, `array.count / length` is zero, so you return zero chunks.  Then, you also return one more chunk of the "leftovers" (in this case, the entire array!) if (`array.count % length > 0` (which you know to be true!) AND `includeRemainder == true` )

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
bobgilmore: So the caller would write `foo.ip_chunks(of: 3)`

Tom-Le: Gotcha!

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
Tom-Le: Got it! I'll push up another commit soon.

brightredchilli:
@elnoss Thanks for the submission - it looks good on this review! And also thanks @bobgilmore for thinking about the performance questions! I think on top of documenting it....it goes without saying that we should write tests for this -

[].ip_chunks(of: 0)
[].ip_chunks(of: 1)
[1,2,3].ip_chunks(of: 1) // this one will ensure we preserve order
[1,2,3].ip_chunks(of: 3) // this one will ensure we preserve order
[1,2,3].ip_chunks(of: 4) // this one will ensure we preserve order

paulrolfe (approved):
:shipit: Looks good to me! What do you think @bobgilmore?

EDIT: Just saw @brightredchilli's comments too. So add those tests in too, of course.

Tom-Le:
Added tests in @brightredchilli's comment. I'm not sure if we need the last two tests though; I think the other tests (on `[0, 1, 2, 3, 4, 5, 6, 7]`) already covered those cases.

SwiftWisdomTests/StandardLibrary/Numbers/Integer+ExtensionsTests.swift
```diff
@@ -26,7 +26,7 @@ class IntegerExtensionTests: XCTestCase {
     }
     
     func testToMagnitudeString() {      
-        let numbers = [14, 1234, 15420003, 5678000000, 8888000000000, 999999, 7666555444333222, 1234567891234567890]
+        let numbers: [Int64] = [14, 1234, 15420003, 5678000000, 8888000000000, 999999, 7666555444333222, 1234567891234567890]
```
brightredchilli, line 29: this snuck in, necessary?

brightredchilli (approved):
Sorry, i really didnt read your test cases properly - the last 2 are actually helpful I think, because it covers empty array case, the [1,2,3].ip_chunks(of: 1) tests are probably less necessary because it is covered by your tests earlier. 

Overall, awesome job!

SwiftWisdomTests/StandardLibrary/Numbers/Integer+ExtensionsTests.swift
Tom-Le: I couldn't get unit tests to compile with Xcode 8.2 when I first checked out the repo. The compiler complained that some of the number literals were too big to fit into normal (32-bit?) Ints.

PR #104, opened 2016-12-28
Remove UIApplication+Extensions to make swift-wisdom extension-safe

mayasaxena:
This file should be removed because app extensions can't access `UIApplication` methods, which makes swift-wisdom unusable in extensions, and also because `openURL` is deprecated in 10.0.

Alternatively, we could create a subspec of swift-wisdom that excludes this file, if anyone is opposed.


Benuuu:
@mayasaxena what's the exact error you're getting when it's compiled together?

mayasaxena:
The error is `'openURL' has been explicitly marked unavailable here`, whether it's the deprecated version or the new one.

alexpersian:
Rather than removing this file and potentially breaking any projects that are currently using `ip_openSettingsApp` I think it would be better to go the subspec route for extensions.

As a plus, we'll have that path to add anything specific for extensions in the future.

edit: the `openURL` deprecation can be handled by updating to the new `openURL:options:completionHandler:` API. We'll need an `if #available(iOS 10, *)` check though.

brightredchilli:
~~@mayasaxena - I would think that it would be up to the developer to not use this in extensions - if you were bitten by it, might be good to write some documentation here. The same issue would happen if the developer wrote `UIApplication.sharedApplication().openXXX...` as opposed to going through this helper method, right? In fact, what is not allowed is the call to sharedApplication, not the call to ip_openSettingsApp~~ oops junk, ignore

mayasaxena:
Sounds good! I'll close this PR and put up a new one with an extension-safe subspec, as well as add some documentation to this file to explain why the method can't be used in an extension and direct them to the extension-safe subspec.

Benuuu:
@mayasaxena would it be possible to have some sort of ifdef guard in this file instead?  Detect whether it is available and compile it if so.

PR #106, opened 2017-01-30
Add custom rx operators ip_delayElements and ip_timeout

Tom-Le:
Add a couple of custom operators to the Rx subspec, plus unit tests.

- `ip_delayElements` to delay elements matching a predicate.
- `ip_timeout` to emit an element whenever a timeout occurs (i.e. whenever observable emits an element, then emits nothing for some time).

I'd love to hear if these are useful additions to swift wisdom, and I could definitely use some help with naming these operators and proofreading their doc comments.

### DIFF
--- a/SwiftWisdom/Rx/Rx+RepeatingTimeouts.swift
+++ b/SwiftWisdom/Rx/Rx+RepeatingTimeouts.swift
@@ -1,0 +1,38 @@
+//
+//  Rx+RepeatingTimeouts.swift
+//  SwiftWisdom
+//
+//  Created by Son Le on 1/30/17.
+//  Copyright © 2017 Intrepid. All rights reserved.
+//
+
+import RxSwift
+
+extension ObservableType {
+
+    /**
+     Creates a new observable sequence that emits the same elements as the source observable, while also emitting the
+     given element whenever the source emits an element, then emits nothing within the specified amount of time.
+
+     Note that unlike `timeout(_,scheduler:)` and `timeout(_,other:scheduler:)`, the returned observable does not
+     terminate after a timeout.
+
+     - parameter dueTime: Maximum duration between emissions before a timeout occurs.
+     - parameter element: Element to be emitted when a timeout occurs.
+     - parameter scehduler: Scheduler to run the timeout timer on.
+     
+     - returns: New observable sequence as described.
+     */
+    public func ip_repeatingTimeouts(
+            interval dueTime: RxTimeInterval,
+            element: E,
+            scheduler: SchedulerType = MainScheduler.instance
+        ) -> Observable<E> {
+        return
+            Observable.of(
+                asObservable(),
+                debounce(dueTime, scheduler: scheduler).map { _ in element }
+            )
+            .merge()
+    }
+}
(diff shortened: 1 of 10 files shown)


vinnyt:
I like it. +1

SwiftWisdomTests/Rx/Rx+DelayElementsTests.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+
+import XCTest
+import RxSwift
+import RxTest
+@testable import SwiftWisdom
+
+final class RxDelayElementsTests: XCTestCase {
+
+    func testDelayingElements() {
```
Benuuu: Could you also add a test of the same sequence with the predicate ($0 % 2 == 0)

It should expect `[1,3,0,2,4]`

Benuuu (approved):

SwiftWisdom/Rx/Rx+Timeout.swift
```diff
@@ -0,0 +1,38 @@
...
+     Note that unlike `timeout(_,scheduler:)` and `timeout(_,other:scheduler:)`, the returned observable does not
+     terminate after a timeout.
+
+     - parameter dueTime: Maximum duration between emissions before a timeout occurs.
+     - parameter element: Element to be emitted when a timeout occurs.
+     - parameter scehduler: Scheduler to run the timeout timer on.
+     
+     - returns: New observable sequence as described.
+     */
+    public func ip_timeout(
```
bobgilmore: I wish that the name (or argument list) made it clear that this repeats, unlike the other `timeout` methods. Maybe something like `repeatingTimeouts`. The plural, especially, would makke that clear.

SwiftWisdomTests/Rx/Rx+DelayElementsTests.swift
Tom-Le: Added another unit test as requested!

SwiftWisdom/Rx/Rx+Timeout.swift
Tom-Le: I renamed it to `ip_repeatingTimeouts(interval:element:scheduler:)`. Does that make it clearer?

paulrolfe (approved):

SwiftWisdom/Rx/Rx+RepeatingTimeouts.swift
```diff
@@ -0,0 +1,38 @@
...
+     Note that unlike `timeout(_,scheduler:)` and `timeout(_,other:scheduler:)`, the returned observable does not
+     terminate after a timeout.
+
+     - parameter dueTime: Maximum duration between emissions before a timeout occurs.
+     - parameter element: Element to be emitted when a timeout occurs.
+     - parameter scehduler: Scheduler to run the timeout timer on.
+     
+     - returns: New observable sequence as described.
+     */
+    public func ip_repeatingTimeouts(
```
bobgilmore, line 26: I like the name change. Renaming the external name of interval to repeatingInterval probably would have worked, too.

PR #107, opened 2017-02-02
Fix bug in which .Finally and .FinallyOn(_) blocks were not being called

cprime:


tomalley104 (approved):

paulrolfe (approved):

paulrolfe:
@cprime Can we get this rebased and in?

PR #108, opened 2017-02-03
DataConvertible protocol, remove NSData references

paulrolfe:
DataConvertible is an easy way to get from number types to Data and back. It could apply to UnsignedIntegers as well, but we already had the capability built in to them without the need for an optional initializer, so UnsignedIntegers actually don't technically conform to the protocol, although they have something very similar.

And I'd also like to remove all the `NSData` conveniences we built. I think everyone should have migrated to using `Data` by now. If this is too extreme to remove so much, I can hold off on that.

Will bump version to `0.7.0`.


SwiftWisdom/Core/Foundation/Data/DataConvertible.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  DataConvertible.swift
+//  SwiftWisdom
+//
+//  Created by Paul Rolfe on 2/3/17.
+//  Copyright © 2017 Intrepid. All rights reserved.
+//
+
```
paulrolfe: I'll add a note that I borrowed this protocol from http://stackoverflow.com/questions/38023838/round-trip-swift-number-types-to-from-data

alexpersian (approved):
Looks good. Might be useful to get a 2nd opinion on the `NSData` removal.

alexpersian (changes requested):
Fix failing tests.

paulrolfe:
retest this please

alexpersian (approved):

bobgilmore:
Are you cool not having a testpoint to exercise `ip_int8Value`? It's a pretty thin wrapper around `.to()` so I wouldn't be *crushed* if it's not exercised, but figured I'd point it out.

brightredchilli:
let's see how this one goes, but in the future, I think we should experiment with deprecating some versions before removing.

brightredchilli (approved):

paulrolfe:
@bobgilmore  yeah might as well test it. Just added that.

PR #109, opened 2017-02-09
Add Multicast protocol to implement multi-delegate systems.

Tom-Le:
Lifted this from a current project 'cause I think it could be useful in other projects. This lets us quickly create classes that can have multiple delegates simply by conforming to `Multicast`.

### DIFF
--- a/SwiftWisdom/Core/CommonTypes/Multicast.swift
+++ b/SwiftWisdom/Core/CommonTypes/Multicast.swift
@@ -1,0 +1,76 @@
+//
+//  Multicast.swift
+//  SwiftWisdom
+//
+//  Created by Logan Wright on 9/4/15.
+//  Copyright © 2015 Intrepid. All rights reserved.
+//
+
+import Foundation
+
+/// Wraps a weak reference to an object.
+/// Useful for storing arrays of references without causing retain cycles.
+public struct Weak<T> where T: AnyObject {
+
+    public weak var weak: T?
+
+    public init(_ weak: T) {
+        self.weak = weak
+    }
+}
+
+/**
+    Conforming to this protocol lets a class implement a multi-delegate system.
+
+    Only subclasses of NSObject can conform to this protocol. In addition, you can only typealias `Delegate` to a
+    class or an `@objc` protocol.
+ */
+public protocol Multicast: class {
+
+    associatedtype Delegate: AnyObject
+
+    var delegateReferences: [Weak<Delegate>] { get set }
+}
+
+public extension Multicast {
+
+    /// Delegates of this class.
+    var delegates: [Delegate] {
+        let existing = delegateReferences.flatMap { $0.weak }
+        if existing.count != delegateReferences.count {
+            delegateReferences = existing.map { Weak($0) }
+        }
+        return existing
+    }
+
+    /**
+        Add a delegate.
+     
+        - parameter delegate: Delegate to be added.
+     */
+    func add(delegate: Delegate) {
+        remove(delegate: delegate)
+        delegateReferences.append(Weak(delegate))
+    }
+
+    /**
+        Remove delegates matching a predicate.
+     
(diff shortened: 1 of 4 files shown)


SwiftWisdom/Core/Multicast.swift
```diff
@@ -0,0 +1,76 @@
+//
+//  Multicast.swift
+//  SwiftWisdom
+//
+//  Created by Logan Wright on 9/4/15.
```
alexpersian: Logan made this?

Tom-Le: Yeah, more or less. I only made slight modifications and added doc comments.

paulrolfe (approved):
:shipit:

SwiftWisdom/Core/CommonTypes/Multicast.swift
```diff
@@ -0,0 +1,76 @@
...
+    public weak var weak: T?
+
+    public init(_ weak: T) {
+        self.weak = weak
+    }
+}
+
+/**
+    Conforming to this protocol lets a class implement a multi-delegate system.
+
```
brightredchilli, line 24: lots of additional spaces between comments...did you mean to leave them in?

SwiftWisdomTests/CommonTypes/MulticastTests.swift
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        let broadcaster = Broadcaster()
+        broadcaster.add(delegate: subscriberA)
+        broadcaster.add(delegate: subscriberB)
+        broadcaster.add(delegate: subscriberC)
+        broadcaster.add(delegate: subscriberA)
+        broadcaster.broadcast(that: "We are intrepid") // get it?
+        if !messageRecipients.ip_contains(all: ["A", "B", "C"]) || messageRecipients.count != 3 {
```
brightredchilli: I think we should prefer the positive assertion here - that messageRecipients contains those items, and the count is three.

SwiftWisdomTests/CommonTypes/MulticastTests.swift
```diff
@@ -0,0 +1,129 @@
...
+            if message == "We are intrepid" {
+                messageRecipients.append("C")
+            }
+        }
+
+        let broadcaster = Broadcaster()
+        broadcaster.add(delegate: subscriberA)
+        broadcaster.add(delegate: subscriberB)
+        broadcaster.add(delegate: subscriberC)
+        broadcaster.add(delegate: subscriberA)
```
brightredchilli, line 93: I would also suggest that this particular test be broken out - this test should be named `testAddOnSameDelegateShouldNotReadd`. 

The overall test should be named `testBroadcastShouldBroadcastToAllDelegates`

SwiftWisdomTests/CommonTypes/MulticastTests.swift
```diff
@@ -0,0 +1,129 @@
...
+                messageRecipients.append("C")
+            }
+        }
+
+        let broadcaster = Broadcaster()
+        broadcaster.add(delegate: subscriberA)
+        broadcaster.add(delegate: subscriberB)
+        broadcaster.add(delegate: subscriberC)
+        broadcaster.removeDelegates { $0 === subscriberA || $0 === subscriberC }
+        broadcaster.broadcast(that: "We are gorgeous")
```
brightredchilli, line 152: love these messages.

brightredchilli (changes requested):

SwiftWisdom/Core/CommonTypes/Multicast.swift
Tom-Le: Yeah, I don't mind deleting them but I thought it looked more readable

brightredchilli (approved):

PR #111, opened 2017-02-17
Updated IntrepidSwiftWisdomTesting for Swift 3.

nservidio:
I updated the version of IntrepidSwiftWisdomTesting.podspec to 0.7.3 because its source is https://github.com/IntrepidPursuits/swift-wisdom.git.
Updated Readme to refer to swift-wisdom instead of intrepid-wisdom for installing testing additions.
Updated IntrepidSwiftWisdom dependency.


IntrepidSwiftWisdomTesting.podspec
```diff
@@ -1,17 +1,17 @@
 Pod::Spec.new do |s|
   s.name          = "IntrepidSwiftWisdomTesting"
-  s.version       = "0.0.7"
+  s.version       = "0.7.3"
```
nservidio: It was a conscious decision to keep this in lockstep with the Intrepid.podspec version. Please let me know if it's better for this to have its own separate version.

brightredchilli:
I would just run it through a test project if you haven't already done so, but it looks good, thanks for taking this on!

brightredchilli (approved):

IntrepidSwiftWisdom.podspec
```diff
@@ -17,6 +17,6 @@ Pod::Spec.new do |s|
 
   s.subspec "Core" do |cs|
     cs.source_files = "SwiftWisdom/Core/**/**/*.swift"
-    cs.dependency 'IP-UIKit-Wisdom', '0.0.9'
+    cs.dependency 'IP-UIKit-Wisdom', '0.0.10'
```
nservidio: Actually, @brightredchilli, if we bump this, we should also bump the IntrepidSwiftWisdom.podspec version, theoretically to 0.7.3...? Since we are no longer really maintaining IntrepidSwiftWisdom, should I just revert this?

cprime:
This pod predates the unified `Intrepid` pod and it's subspecs. It seems like this would be a good time to migrate this into being a `Intrepid` subspec too.

IntrepidSwiftWisdom.podspec
brightredchilli: no harm to bump both if you're updating. however, none of the changes will be visible until we make a push with the proper tags.

bobgilmore (approved):

SwiftWisdomTests/XCTestTests.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  XCTestTests.swift
+//  SwiftWisdom
+//
+//  Created by Colden Prime on 4/12/17.
+//  Copyright © 2017 Intrepid. All rights reserved.
+//
+
+//ip_XCTAssertEqualOptional
```
nservidio: Did you mean to leave these here?

IntrepidSwiftWisdom.podspec
```diff
@@ -1,22 +0,0 @@
-Pod::Spec.new do |s|
-  s.name          = "IntrepidSwiftWisdom"
```
nservidio, line 2: It looks like a few projects use IntrepidSwiftWisdom. Is it ok to remove this spec?

SwiftWisdomTests/XCTestTests.swift
```diff
@@ -0,0 +1,28 @@
...
+
+        ip_XCTAssertEqualOptional(value1, value2)
+    }
+
+    func testAssertEqualWithAccuracyOptional() {
+        let value1: Float? = 1
+        let value2: Float? = 1.00001
+
+        ip_XCTAssertEqualWithAccuracyOptional(value1, value2, accuracy: 0.0001)
+    }
```
nservidio: Can you test to make that these assert if you pass erroneous values?

IntrepidSwiftWisdom.podspec
cprime: I think it is safe to remove this podspec because:
- the IntrepidSwiftWisdom spec still exists in the the spec repo
- the old tags still exist on https://github.com/IntrepidPursuits/swift-wisdom that reference IntrepidSwiftWisdom

PR #113, opened 2017-02-20
Convert between (Int|UInt)(16|32|64) and Data while respecting endianness

cprime:


SwiftWisdomTests/Data/DataConversionTests.swift
```diff
@@ -50,7 +50,25 @@ class DataConversionTests: XCTestCase {
         let data = Data(byte: 42)
         XCTAssert(data.ip_uint8Value == 42)
     }
-    
+
+    func testUInt16() {
+        let data = Data(bytes: [0x00, 0x01])
+        XCTAssert(data.ip_bigEndianUInt16Value == 1)
+        XCTAssert(data.ip_littleEndianUInt16Value == UInt16(1 << 8))
```
paulrolfe: Can we also test when a nil should be returned in these?

paulrolfe (changes requested):
I like it. Not necessary for this PR, but maybe it should  also exist on signed Int16, 32, 64?

cprime:
@paulrolfe I've added Int16, 32, 64 support as well

paulrolfe (approved):

paulrolfe:
retest this please

cprime:
retest this please

paulrolfe:
Looks like some tests need to be fixed.

SwiftWisdom/Core/StandardLibrary/Numbers/Integer+Extensions.swift
```diff
@@ -56,3 +56,32 @@ extension NumberFormatter {
         return decimalFormatter
     }
 }
+
+extension Integer {
+    fileprivate var ip_data: Data {
```
brightredchilli, line 61: I would rather see this as a Data constructor.

brightredchilli (approved):

brightredchilli:
what was the change, ps?

SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
```diff
@@ -20,7 +20,8 @@ extension Data {
     // From here: http://stackoverflow.com/questions/38023838/round-trip-swift-number-types-to-from-data
     public func to<T>(type: T.Type) -> T? {
         guard self.count > 0 else { return nil }
-        return self.withUnsafeBytes { $0.pointee }
```
cprime, line 23: These changes fixed the non-deterministic test failures

SwiftWisdom/Core/StandardLibrary/Numbers/Integer+Extensions.swift
cprime: This is more inline with the `var ip_data: Data` on all of the other primitive types

SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
cprime: @brightredchilli

cprime: no idea why

cprime:
retest this please

cprime:
retest this please

PR #115, opened 2017-03-02
Various Tests

ButkiewiczP:
No added functionality -- Just various tests for Core classes in an effort to raise code coverage.

### DIFF
--- a/SwiftWisdomTests/ColorDescriptorTests.swift
+++ b/SwiftWisdomTests/ColorDescriptorTests.swift
@@ -50,22 +50,22 @@ class ColorDescriptorTests: XCTestCase {
     }
     
     func testUnicodeScalarColor() {
-        let color: ColorDescriptor = ColorDescriptor.init(unicodeScalarLiteral: "#0000FF")
+        let color: ColorDescriptor = ColorDescriptor(unicodeScalarLiteral: "#0000FF")
         XCTAssert(color.color == UIColor(red: 0, green: 0, blue: 1, alpha: 1))
     }
     
     func testStringLiteralColor() {
-        let color: ColorDescriptor = ColorDescriptor.init(stringLiteral: "#0000FF")
+        let color: ColorDescriptor = ColorDescriptor(stringLiteral: "#0000FF")
         XCTAssert(color.color == UIColor(red: 0, green: 0, blue: 1, alpha: 1))
     }
     
     func testExtendedGraphemeClusterLiteralColor() {
-        let color: ColorDescriptor = ColorDescriptor.init(extendedGraphemeClusterLiteral: "#0000FF")
+        let color: ColorDescriptor = ColorDescriptor(extendedGraphemeClusterLiteral: "#0000FF")
         XCTAssert(color.color == UIColor(red: 0, green: 0, blue: 1, alpha: 1))
     }
     
     func testRawValueColor() {
-        let color: ColorDescriptor? = ColorDescriptor.init(rawValue: "#FF0000")
+        let color: ColorDescriptor? = ColorDescriptor(rawValue: "#FF0000")
         XCTAssertNotNil(color)
         XCTAssert(color?.color == UIColor(red: 1, green: 0, blue: 0, alpha: 1))
     }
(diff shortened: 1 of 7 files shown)


SwiftWisdomTests/ColorDescriptorTests.swift
```diff
@@ -21,4 +22,51 @@ class ColorDescriptorTests: XCTestCase {
...
+        let rhsHexDescriptor: ColorDescriptor = "#FF0000"
+        XCTAssertFalse(lhsHexDescriptor == rhsHexDescriptor)
+        
+        let whiteHexDescriptor: ColorDescriptor = "#FFFFFF"
+        let whiteRGBHexDescriptor: ColorDescriptor = "255,255,255,255"
+        XCTAssertFalse(whiteHexDescriptor == whiteRGBHexDescriptor)
+    }
+    
+    func testUnicodeScalarColor() {
+        let color: ColorDescriptor = ColorDescriptor.init(unicodeScalarLiteral: "#0000FF")
```
brightredchilli: s/ColorDescriptor.init(...)/ColorDescriptor(..) 
s/color: ColorDescriptor//

brightredchilli: same for below

SwiftWisdomTests/CommonTypes/ResultTests.swift
```diff
@@ -0,0 +1,52 @@
...
+        XCTAssertFalse(failureResult.isSuccess)
+        XCTAssert(failureResult.isFailure)
+    }
+    
+    func testFailureResultStorage() {
+        XCTAssertNil(failureResult.value)
+        do {
+          let _ = try failureResult.extract()
+        } catch {
+            return XCTAssertNotNil(error)
```
brightredchilli: seems to be better to more explicitly use XCTAssertThrows

SwiftWisdomTests/StandardLibrary/Date/TimeOfDayTests.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+//  Created by Patrick Butkiewicz on 3/1/17.
+//  Copyright © 2017 Intrepid. All rights reserved.
+//
+
+import Foundation
+import XCTest
+import SwiftWisdom
+
+final class TimeOfDayTests: XCTestCase {
```
brightredchilli, line 13: these tests will fail when run on non EST machine. Either inject the date formatter, or query the current timezone in the test and calculate based on offsets.

SwiftWisdomTests/StandardLibrary/Set/SetTests.swift
```diff
@@ -0,0 +1,31 @@
...
+//  SwiftWisdom
+//
+//  Created by Patrick Butkiewicz on 3/1/17.
+//  Copyright © 2017 Intrepid. All rights reserved.
+//
+
+import Foundation
+import XCTest
+
+final class SetTests: XCTestCase {
```
brightredchilli, line 12: i think these tests are solid, but we should deprecate these. Set now implements Collection, and ExpressibleByArrayLiteral. Maybe the only not explicitly defined operator is ip_passes...be open to thoughts on this one.

SwiftWisdomTests/StandardLibrary/Date/TimeOfDayTests.swift
ButkiewiczP: Great catch. Updated tests in latest commit to be independent of time zones.

SwiftWisdomTests/StandardLibrary/Set/SetTests.swift
ButkiewiczP: I'm all for deprecating these. While they're still around it would be nice to have tests for them, but I'm all proposing to deprecate these as part of another PR or whatever

SwiftWisdomTests/StandardLibrary/Date/NSDateComparableTests.swift
```diff
@@ -0,0 +1,34 @@
...
+import Foundation
+import XCTest
+
+final class NSDateComparableTests: XCTestCase {
+    let olderDate = NSDate(timeIntervalSince1970: 0)
+    let olderDateCopy = NSDate(timeIntervalSince1970: 0)
+    let moreRecentDate = NSDate(timeIntervalSinceNow: 0)
+    
+    func testLessThanDate() {
+        XCTAssert(olderDate < moreRecentDate)
```
brightredchilli, line 18: xctassertfalse(olderDate < olderDateCopy) ..  dont exactly think this should fail but...

brightredchilli (approved):

SwiftWisdomTests/StandardLibrary/Set/SetTests.swift
brightredchilli: ok, agreed

PR #119, opened 2017-04-04
ip_safely subscripting strings

paulrolfe:
Returns a substring in the given range.
If the range is beyond the string's length, returns a substring up to the bounds of the string.
For example:
```
let hello = "Hello"
hello[ip_safely: 1...5] // "ello"
hello[ip_safely: -1...3] // "Hell"
```

### DIFF
--- a/SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
+++ b/SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
@@ -15,21 +15,31 @@ extension String {
     }
 
     /// Returns a substring in the given range.
-    /// If the range is beyond the string's length, returns the original string.
+    /// If the range is beyond the string's length, returns the substring up to it's bounds.
     public subscript(ip_safely range: Range<Int>) -> String {
         let chars = Array(characters)
-        guard range.upperBound < chars.count else { return self }
-        let substringCharacters = chars[range]
-        return String(substringCharacters)
+        if range.lowerBound < 0 {
+            return self[ip_safely: 0..<range.upperBound]
+        } else if range.upperBound > chars.count {
+            let newRange = range.lowerBound..<chars.count
+            return String(chars[newRange])
+        } else {
+            return String(chars[range])
+        }
     }
 
     /// Returns a substring in the given range.
-    /// If the range is beyond the string's length, returns the original string.
+    /// If the range is beyond the string's length, returns the substring up to it's bounds.
     public subscript(ip_safely range: CountableClosedRange<Int>) -> String {
         let chars = Array(characters)
-        guard range.upperBound < chars.count else { return self }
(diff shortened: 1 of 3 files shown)


alexpersian (approved):
Good addition.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -14,6 +14,24 @@ extension String {
         return String(substringCharacters)
     }
 
+    /// Returns a substring in the given range.
+    /// If the range is beyond the string's length, returns the original string.
```
bobgilmore: Does it make sense to return the original string if the `Range` extends past the end of the string but the lower bound isn't zero?  For example, shouldn't

    let hello = "Hello"
    hello[ip_safely: 1...5]

return `ello` rather than `Hello`?

cprime (approved):

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
paulrolfe: Oops, yes, I actually had it this way and then changed it. Now I remember why I did it like that! Good catch.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -15,21 +15,31 @@ extension String {
     }
 
     /// Returns a substring in the given range.
-    /// If the range is beyond the string's length, returns the original string.
+    /// If the range is beyond the string's length, returns the substring up to it's bounds.
```
bobgilmore: Typo on it's vs it's.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -14,6 +14,34 @@ extension String {
         return String(substringCharacters)
     }
 
+    /// Returns a substring in the given range.
+    /// If the range is beyond the string's length, returns the substring up to it's bounds.
```
bobgilmore, line 18: Typo it's vs it's

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -14,6 +14,34 @@ extension String {
...
+        } else if range.upperBound > chars.count {
+            let newRange = range.lowerBound..<chars.count
+            return String(chars[newRange])
+        } else {
+            return String(chars[range])
+        }
+    }
+
+    /// Returns a substring in the given range.
+    /// If the range is beyond the string's length, returns the substring up to it's bounds.
```
bobgilmore, line 32: Typo it's vs its

PR #122, opened 2017-04-10
DataConvertible fix for 9.3

paulrolfe:
As of Swift 3.1, the conversion from Data to Int64 broke on iOS 9.3 (http://stackoverflow.com/questions/43099577/withunsafebytes-with-swift-3-1-and-ios-9-3)

- I've addressed this by adding on the empty bytes manually during the conversion process. And funneling all Data to Number conversions through DataConvertible.
- Also fixed one last `.pi`

### DIFF
--- a/SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
+++ b/SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
@@ -16,12 +16,13 @@ extension Data {
         guard self.count > 0 else { return nil }
         return map { String(format: "%02hhx", $0) }.joined()
     }
+
+    // MARK: Convenience
     
     // From here: http://stackoverflow.com/questions/38023838/round-trip-swift-number-types-to-from-data
-    public func to<T>(type: T.Type) -> T? {
+    public func to<T: DataConvertible>(type: T.Type) -> T? {
         guard self.count > 0 else { return nil }
-        let value: T = self.withUnsafeBytes { $0.pointee }
-        return value
+        return T(ip_data: self)
     }
     
     public var ip_hexInt: UInt? {
--- a/SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
+++ b/SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
@@ -19,7 +19,6 @@ extension Data {
 
     // MARK: Convenience
     
-    // From here: http://stackoverflow.com/questions/38023838/round-trip-swift-number-types-to-from-data
     public func to<T: DataConvertible>(type: T.Type) -> T? {
         guard self.count > 0 else { return nil }
         return T(ip_data: self)
--- a/SwiftWisdom/Core/Foundation/Data/DataConvertible.swift
+++ b/SwiftWisdom/Core/Foundation/Data/DataConvertible.swift
@@ -18,13 +18,24 @@ public protocol DataConvertible {
 extension DataConvertible {
     
     public init?(ip_data data: Data) {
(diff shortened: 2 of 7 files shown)


tzm41:
Thanks for fixing the last `.pi`.

alexpersian (approved):

SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
```diff
@@ -16,12 +16,13 @@ extension Data {
         guard self.count > 0 else { return nil }
         return map { String(format: "%02hhx", $0) }.joined()
     }
+
+    // MARK: Convenience
     
     // From here: http://stackoverflow.com/questions/38023838/round-trip-swift-number-types-to-from-data
```
Benuuu: Is this still relevant?

SwiftWisdom/Core/Foundation/Data/DataConvertible.swift
```diff
@@ -18,13 +18,24 @@ public protocol DataConvertible {
...
     
     public init?(ip_data data: Data) {
-        guard data.count == MemoryLayout<Self>.size else { return nil }
-        self = data.withUnsafeBytes { $0.pointee }
+        let size = MemoryLayout<Self>.size
+        if data.count < size {
+            let difference = size - data.count
+            var modifiedData = data
+            for _ in 0..<difference {
+                modifiedData.append(0)
```
Benuuu: Does anyone know if appending an array of size `difference` would be faster than iterating over `difference` and appending one by one?

`modifiedData.append(contentsOf: Array<UInt8>(count:difference, repeatedValue: 0))`

SwiftWisdomTests/BitMaskTests.swift
```diff
@@ -34,9 +34,9 @@ class BitMaskTests: XCTestCase {
     }
     
     func testDataInitializer() {
-        let data = "0x1f".ip_dataFromHexadecimalString()!
-        let mask = UInt8(ip_data: data)
-        let hex = data.ip_hexInt!
+        let data = "1f".ip_dataFromHexadecimalString()!
+        let mask = UInt8(ip_data: data)!
```
Benuuu, line 38: Lotta force unwrapping here, could we have a guard instead that fails with XCTAssertFail or something?  What are people's general thoughts on this in tests?

paulrolfe: I've always thought force unwrapping was ok in tests, because any nil value unwrapped would act like a failure of the tests.

SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
paulrolfe: Yes, but it is already in the DataConvertible protocol where it belongs, so we can remove it from here.

SwiftWisdomTests/BitMaskTests.swift
Benuuu: Do they crash the test suite or does it fail and continue on with other tests?

Benuuu (approved):

SwiftWisdom/Core/Foundation/Data/DataConvertible.swift
paulrolfe: I like that better. Not sure there's any performance difference.

SwiftWisdomTests/BitMaskTests.swift
paulrolfe: It's a crash

PR #124, opened 2017-04-19
Added missing double bind methods from RxExamples

djbrooks111:
These two methods are in https://github.com/ReactiveX/RxSwift/blob/master/RxExample/RxExample/Operators.swift, but were missing from Wisdom.

This solves an error that Xcode would throw in the following example:
```
@IBOutlet weak var textView: UITextView!
var caption = Variable("")
...
textView.rx.text <-> caption // This is where Xcode would throw a "Cannot convert type" error
```


tzm41:
Use `textView.rx.text.orEmpty <-> caption`.

Benuuu (commented):
Please write a basic test.

djbrooks111:
@tzm41's suggestion worked. @Benuuu do we still want this PR then?

Benuuu:
Lets drop it then

PR #126, opened 2017-04-20
Changes array's safe subscript to ip_safely

alexpersian:
* Brings array's safe subscript naming convention into line with the others.


paulrolfe:
I like it, we should definitely bring this in-line with all the other `ip_safely` uses. However this will certainly break apps using this subscripting. So I'd like to get some more buy in from everyone before making this breaking change. @cprime @vinnyt  @markdintrepid @aspitz-intrepid

cprime (approved):

brightredchilli:
I think since we're changing method signature, we should deprecate the old, and add a new method with documentation..

paulrolfe (approved):

SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -114,15 +114,30 @@ extension Array where Element: Hashable {
...
-    /// Provides a way to safely index into an array. If the index is beyond the array's final element this method will return `nil`.
+    /// DEPRECATED: Use `ip_safely` instead.
+    ///
+    /// Provides a way to safely index into an array. If the index is beyond the array's 
+    /// final element this method will return `nil`.
     ///
     /// - parameter safe: Index of the element to return
     ///
     /// - returns: An `Element` if the index was correct, or `nil` if it goes beyond the array
+    @available(*, deprecated: 0.8.0, message: "Use `ip_safely` instead.")
```
paulrolfe: Since we're going to keep this, seems like we could just bump the version to `0.7.11`

cprime (approved):

PR #129, opened 2017-06-08
Swift 4 Conversion

tzm41:
Not intended to be production ready yet, but to kickstart the process now!
- Pod install using latest CocoaPods version
- Conform to Apple style guide to move from `(Void)` to `()`
- Annotate Obj-C dynamic methods
- New associated type `Element` added to `Collection` type
- `Integer` protocol has been removed
- `IntMax` type alias has been removed
- Some types conformances have been built into the types
- Other misc changes


SwiftWisdom/Core/StandardLibrary/Array/Array+Utilities.swift
```diff
@@ -154,9 +154,9 @@ extension Collection {
     /// If there's an odd number the return array is just one element.
     /// If there are an even number it will return the two middle elements.
     /// The two middle elements will be flipped if the array has an even number.
-    public var ip_middleElements: [Iterator.Element] {
+    public var ip_middleElements: [Element] {
         guard count > 0 else { return [] }
-        let needsAverageOfTwo = count.toIntMax().ip_isEven
+        let needsAverageOfTwo = Int(count).ip_isEven
```
chow-land, line 159: isn't 'count' already an Int?

tzm41: No...

tzm41: It's an `IndexDistance`

tzm41:
Some protocols and code seem to become obsolete.. For example, `Mathable`. We might want to reconsider the use cases for those code.

tzm41:
https://github.com/apple/swift-evolution/blob/master/proposals/0104-improved-integers.md

tzm41:
Need to find a way to build this PR using Xcode 9.

PR #145, opened 2017-10-10
Addressing namespace collisions & String.characters deprecation

aspitz-intrepid:
Updated Podfile. Updated `Delegate` to `MulticastDelegate` to avoid namespace collision issues. Addressed `String.characters` being deprecated

### DIFF
--- a/SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
+++ b/SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
@@ -3,48 +3,55 @@ import Foundation
 extension String {
 
     public subscript(range: Range<Int>) -> String {
-        let chars = Array(characters)
-        let substringCharacters = chars[range]
-        return String(substringCharacters)
+        let lowerBound = index(startIndex, offsetBy: range.lowerBound)
+        let upperBound = index(startIndex, offsetBy: range.upperBound)
+        let indexRange = lowerBound..<upperBound
+        return String(self[indexRange])
+    }
+
+    public subscript(range: CountableRange<Int>) -> String {
+        let lowerBound = index(startIndex, offsetBy: range.lowerBound)
+        let upperBound = index(startIndex, offsetBy: range.upperBound)
+        let indexRange = lowerBound..<upperBound
+        return String(self[indexRange])
     }
 
     public subscript(range: CountableClosedRange<Int>) -> String {
-        let chars = Array(characters)
-        let substringCharacters = chars[range]
-        return String(substringCharacters)
+        let lowerBound = index(startIndex, offsetBy: range.lowerBound)
+        let upperBound = index(startIndex, offsetBy: range.upperBound)
+        let indexRange = lowerBound...upperBound
+        return String(self[indexRange])
     }
 
     /// Returns a substring in the given range.
     /// If the range is beyond the string's length, returns the substring up to it's bounds.
     public subscript(ip_safely range: Range<Int>) -> String {
-        let chars = Array(characters)
         if range.lowerBound < 0 {
             return self[ip_safely: 0..<range.upperBound]
-        } else if range.upperBound > chars.count {
-            let newRange = range.lowerBound..<chars.count
-            return String(chars[newRange])
+        } else if range.upperBound > self.count {
+            let newRange = range.lowerBound..<self.count
+            return String(self[newRange])
         } else {
-            return String(chars[range])
+            return String(self[range])
         }
     }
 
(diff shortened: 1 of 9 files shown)


SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -3,48 +3,55 @@ import Foundation
 extension String {
 
     public subscript(range: Range<Int>) -> String {
-        let chars = Array(characters)
-        let substringCharacters = chars[range]
-        return String(substringCharacters)
+        let lowerBound = index(startIndex, offsetBy: range.lowerBound)
```
bobgilmore, line 6: In Swift 4, can't this entire method simply be implemented as
`return String(self[range])`? And similarly for some of the other methods?

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -3,48 +3,55 @@ import Foundation
...
+        let upperBound = index(startIndex, offsetBy: range.upperBound)
+        let indexRange = lowerBound...upperBound
+        return String(self[indexRange])
     }
 
     /// Returns a substring in the given range.
     /// If the range is beyond the string's length, returns the substring up to it's bounds.
```
bobgilmore: Typo in the doc comment: `its` not `it's`

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -3,48 +3,55 @@ import Foundation
...
     public subscript(ip_safely range: Range<Int>) -> String {
-        let chars = Array(characters)
         if range.lowerBound < 0 {
             return self[ip_safely: 0..<range.upperBound]
-        } else if range.upperBound > chars.count {
-            let newRange = range.lowerBound..<chars.count
-            return String(chars[newRange])
+        } else if range.upperBound > self.count {
```
bobgilmore: I think that we're missing a case here (even in the original implementation); if `range.lowerBound < 0` *and* `range.upperBound > self.count`, this will fail.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -3,48 +3,55 @@ import Foundation
...
+            let newRange = range.lowerBound..<self.count
```
bobgilmore: Unnecessary `self`s  here and throughout. The `self.count`s can simply be `count`s

SwiftWisdom/Core/StandardLibrary/String/String+Numbers.swift
```diff
@@ -22,6 +22,6 @@ extension String {
     }
 
     public var ip_length: Int {
-        return self.characters.count
+        return self.count
```
bobgilmore: Unnecessary `self`.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
aspitz-intrepid: unfortunately no. you can use subscript range between String.index ranges. The other methods in this class actually use this class :)

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
aspitz-intrepid: if the `lowerBound < 0` then the code calls itself but passing in 0 as the lower bound. this second call should pass the lower bounds test and then tests the `upperBound`, building a new range.

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
aspitz-intrepid: :+1:

SwiftWisdom/Core/StandardLibrary/String/String+Numbers.swift
aspitz-intrepid: :+1:

ghost:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 167 tests, with 0 failures (0 unexpected) in 0.630 (0.710) seconds</td>
    </tr>
  </tbody>
</table>

</p>

PR #149, opened 2018-03-13
Check if already on main thread

MaxHasADHD:
Don't need to dispatch async if already on the main thread.


ghost:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 169 tests, with 0 failures (0 unexpected) in 0.467 (0.544) seconds</td>
    </tr>
  </tbody>
</table>

</p>

bobgilmore:
I'm not behind this change. If someone is on the main thread already, the current implementation of `Main` will (I believe) enqueue the `function` call onto the *end* of the main thread. That would defer execution of `function` until some *future* run through the event queue. This may well be what the caller intended. It's what I've always assumed.

This change would break that behavior, resulting in `function` being executed immediately and synchronously.

If the caller wants this kind of conditional behavior, I think they should perform the check at the call site.  If we really want to provide this behavior in a library call, I advocate for another method, or (at least) have an optional boolean parameter on *this* method to specify the behavior.

PR #147, opened 2017-11-03
SafeArea helpers

alexpersian:
In iOS 11.0 the new `safeAreaInsets` and `safeAreaLayoutGuide` features were added to handle laying out based on the Safe Area. This feature is not available in iOS 10 and below though, and if you try to use them the layout will go to the `superview` instead of the layout guides. Meaning that you need to make sure to wrap the safe area access in `#available` calls all the time.

This PR adds helper properties to remove the need for wrapping iOS 11.0 checks when setting layout guides.

PR #150, opened 2018-04-13
Alex/update to fix new xcode issues

alexpersian:
Multiple issues were found when attempting to update the pod trunk after merging in PR #147.

These issues were caused by a few things:
- flatMap needed to switch to compactMap in three places (ref: https://github.com/apple/swift-evolution/blob/master/proposals/0187-introduce-filtermap.md)
- The podspec still showed support for iOS 8.0 which prevented the use of NSLayoutAnchors (pod spec linting failed)
- Our project file still showed support for Swift 2.3 in the main proj file, while the individual targets were flagged for 4.1 already.

The 0.9.2 tag has not been released to Cocoapods, but exists in the repo. This tag currently points to a broken version should anyone manually install the project and use Xcode 9.3 (Swift 4.1). After the PR is merged should the tag be rewritten to point to this commit (overwriting 0.9.2), or should that tag stay around?

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.9.3"
+  s.version       = "0.10.0"
   s.summary       = "Swift Bag"
   s.description   = <<-DESC
                     Collection of extensions and utility classes by and for the developers at Intrepid Pursuits.
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.9.2"
+  s.version       = "0.9.3"
   s.summary       = "Swift Bag"
   s.description   = <<-DESC
                     Collection of extensions and utility classes by and for the developers at Intrepid Pursuits.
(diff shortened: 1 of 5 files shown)


eltonium (approved):
LGTM

cprime (approved):

Intrepid.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.9.2"
+  s.version       = "0.9.3"
```
paulrolfe: We might want to go to `0.10.0` since this is a breaking change that will no longer compile in xcode 9.2 <

Intrepid.podspec
```diff
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/IntrepidPursuits/swift-wisdom.git", :tag => "#{s.version}" }
   s.exclude_files = "tests/**/*"
   s.platform      = :ios
-  s.ios.deployment_target = "8.0"
+  s.ios.deployment_target = "9.0"
```
paulrolfe, line 28: Does this have to change? I don't think we should bump target requirements unless required by the API's we're using. Seems like `compactMap` works on 8.0, still as long as it's compiled in xcode 9.3.

paulrolfe (changes requested):

Intrepid.podspec
alexpersian: The spec bump was more to handle the `NSLayoutAnchors` usage that was introduced through PR #147 

The project itself targets iOS 9.0, so the issue wasn't found until the Podspec was linted before attempting to push to the trunk.

ghost:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 169 tests, with 0 failures (0 unexpected) in 0.565 (0.696) seconds</td>
    </tr>
  </tbody>
</table>

</p>

bobgilmore (approved):

Intrepid.podspec
paulrolfe: Ok, understood ⚓️

paulrolfe (approved):

PR #153, opened 2018-06-11
Updated SwiftWisdom for Swift 4.2

aspitz-intrepid:
- Cleaned up `subscript` `CountableRange` issue
- Updated version info

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.10.0"
+  s.version       = "0.11.0"
   s.summary       = "Swift Bag"
   s.description   = <<-DESC
                     Collection of extensions and utility classes by and for the developers at Intrepid Pursuits.
@@ -25,8 +25,8 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/IntrepidPursuits/swift-wisdom.git", :tag => "#{s.version}" }
   s.exclude_files = "tests/**/*"
   s.platform      = :ios
-  s.ios.deployment_target = "9.0"
-  s.pod_target_xcconfig = { 'SWIFT_VERSION' => '4.0' }
+  s.ios.deployment_target = "10.0"
+  s.pod_target_xcconfig = { 'SWIFT_VERSION' => '4.2' }
   s.default_subspec = "Core"
 
   s.subspec "Core" do |cs|
--- a/SwiftWisdom.xcodeproj/project.pbxproj
+++ b/SwiftWisdom.xcodeproj/project.pbxproj
@@ -809,7 +809,6 @@
 				80232B9C1BF2EE2E00818B6E /* Frameworks */,
 				80232B9D1BF2EE2E00818B6E /* Resources */,
 				BEDBF20DB05D76F60C858449 /* [CP] Embed Pods Frameworks */,
-				E3F5EA111B900D6479C8D173 /* [CP] Copy Pods Resources */,
 				3169F10A1F06850F00C917B2 /* SwiftLint */,
 			);
 			buildRules = (
@@ -830,7 +829,6 @@
 				80232BB01BF2EE2E00818B6E /* Frameworks */,
 				80232BB11BF2EE2E00818B6E /* Resources */,
 				FAA5EC81369AE73A0E00ED4B /* [CP] Embed Pods Frameworks */,
-				77882022D65A5916A7FFD999 /* [CP] Copy Pods Resources */,
 			);
 			buildRules = (
 			);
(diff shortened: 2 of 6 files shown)


Intrepid.podspec
```diff
@@ -25,8 +25,8 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/IntrepidPursuits/swift-wisdom.git", :tag => "#{s.version}" }
   s.exclude_files = "tests/**/*"
   s.platform      = :ios
-  s.ios.deployment_target = "9.0"
```
alexpersian, line 28: We just moved this up to 9.0 from 8.0, do we really want to drop 9.0 support already?

SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -854,6 +852,7 @@
 				TargetAttributes = {
 					80232B9E1BF2EE2E00818B6E = {
 						CreatedOnToolsVersion = 7.1;
+						DevelopmentTeam = THP5EV5EJ9;
```
alexpersian, line 855: Might want to drop the dev team inclusion here.

SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -1289,6 +1258,7 @@
 			buildSettings = {
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = "$(inherited)";
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				DEVELOPMENT_TEAM = THP5EV5EJ9;
```
alexpersian, line 1261: Ditto on dev team

SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -1307,6 +1277,7 @@
...
```
alexpersian, line 1280: Ditto on dev team

SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
```diff
@@ -99,11 +99,6 @@ extension Data {
         return self[ip_safely: idx...idx]
     }
 
-    /// Safer version of `subdata`. Will return nil if the range is outside the bounds of the data.
-    public subscript(ip_safely range: CountableRange<Int>) -> Data? {
```
alexpersian, line 103: Was `CountableRange` just removed in 4.2?

I'm not finding any information around that online.

alexpersian (changes requested):

SwiftWisdom/Core/Foundation/Data/Data+Extensions.swift
aspitz-intrepid: See https://github.com/apple/swift/blob/master/stdlib/public/core/Range.swift#L287

alexpersian: This would still be needed for Swift < 4.2 though, no?

aspitz-intrepid: The problem is that I can't seem to find a good way to only build this code if you're NOT on Swift 4.2

Bob-Gilmore-Intrepid: Would this work?
```swift
#if swift(>=4.2)
#else
    /// Safer version of `subdata`. Will return nil if the range is outside the bounds of the data.
    public subscript(ip_safely range: CountableRange<Int>) -> Data? {
        return self[ip_safely: Range<Int>(range)]
    }
#endif
```

bobgilmore: Would this work?
```swift
#if swift(>=4.2)
#else
    /// Safer version of `subdata`. Will return nil if the range is outside the bounds of the data.
    public subscript(ip_safely range: CountableRange<Int>) -> Data? {
        return self[ip_safely: Range<Int>(range)]
    }
#endif
```

alexpersian: @bobgilmore nope, because the compiler version in Xcode 10 still flags it as an invalid redeclaration.

Until this proposal is implemented in one of the 4.2 releases I think we should hold off on this PR.
https://github.com/apple/swift-evolution/blob/master/proposals/0212-compiler-version-directive.md

bobgilmore: This is implemented in the Swift version that ships with Xcode 10 beta 4. It's probably time to try again.

paulrolfe:
I think we're using #155 instead of this... closing 🤞

PR #157, opened 2018-10-02
Add a Core-App-Extension-Safe subspec

bobgilmore:
NOTE: I'm not certain if THIS is the right thing to do (leave the default spec as EXTENSION-UNSAFE and add a subspec that's extension-safe), 

OR if we should make the default spec EXTENSION-SAFE and add a subspecies to add the UIApplication bits, similar to how people have to add the Rx-related bits separately. This would be a very small breaking change, so we might want it to go into the 0.11.0 release.

-------------

The code in UIApplication+Extensions.swift (with implicit references
to Application.shared) is not safe for use in a Today Widget.
Including it in a project with widgets generates...

EITHER the following compile-time warning:

ld: warning: linking against a dylib which is not safe for use in
application extensions: <path follows>

OR a compile time-error that the methods are marked unavailable,
depending on how Cocoapods generates the project files, especially
with regard to the "Require Only App-Extension-Safe API" setting.

A project has been creating a subspec in a separate branch, but as
it's of general utility, I'd like to move the subspec into the main
branch.

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -34,6 +34,12 @@ Pod::Spec.new do |s|
     cs.dependency 'IP-UIKit-Wisdom', '0.0.10'
   end
 
+  s.subspec "Core-App-Extension-Safe" do |ext|
+    ext.source_files = "SwiftWisdom/Core/**/**/*.swift"
+    ext.exclude_files = "SwiftWisdom/Core/UIKit/UIApplication+Extensions.swift"
+    ext.dependency 'IP-UIKit-Wisdom', '0.0.10'
+  end
+
   s.subspec "Rx" do |rx|
     rx.source_files = "SwiftWisdom/Rx/**/**/*.swift"
     rx.dependency 'RxSwift', '~> 4.0'


MaxHasADHD (approved):

Benuuu:
Related #104

paulrolfe (changes requested):
Offline we decided to go with @bobgilmore's second proposal.
```
make the default spec EXTENSION-SAFE and add a subspecies to add the UIApplication bits
```

bobgilmore:
I’ve done some investigation. As far as I can tell, the separate specs (such as the various `stetson` branches maintain) are no longer needed.  Two things:
1. I’ve picked up vague hints that this was an iOS 8.0 and earlier SDK issue, and
2. On two separate occasions, months apart, I’ve successfully built my app AND its Today Extension using the “main” swift-wisdom branch. Hacks had to be made, because of some issues I won’t go into here, but the UIApplication / openURL issue was NOT one of them.

So my plan (for now) is to close this PR. If I’m way off base, I can always resurrect it.

PR #159, opened 2018-10-31
SW – adding framework targets to `SwiftWisdom.xcodeproj` so that `IpS…

Steven-Wu-ConnectedLab:
…wiftWisdom` and `RxIpSwiftWisdom` can be consumed via Carthage

**DO NOTE MERGE YET**

### Changes

- Added build targets `IpSwiftWisdom` and `RxIpSwiftWisdom`.
- Removed Pod references.
- Added `Cartfile`.
- Update .gitignore.
- Wired up framework dependencies for all build targets.
- Ensured that all existing targets build and that tests run.
- Updated README

### Dependent Work

This PR is dependent on https://github.com/IntrepidPursuits/uikit-wisdom/pull/18 .  After Ip-UI-Kit-Wisdom is consumable via Carthage, the `Cartfile` on this repo will need to be updated to point to the `IntrepidPursuits/uikit-wisdom` instead.

### Notes

- Post-install steps in the `Podfile` seem to mainly deal with Swift 4.2 migration which look to be complete now in `0.11.0`, so I did not make project or script changes to accommodate those custom build configurations.
- Podspec name is 'Intrepid', which seems inconsistent with name of repo 'swift-wisdom'.  When creating the new targets, I chose to use the names `IpSwiftWisdom` and `RxIpSwiftWisdom` to align with the repo name.  May want to consider changing the podspec name to match since it means that folks who consume via Cocoapods are currently doing `import Intrepid` as opposed to `import IpSwiftWisdom`.
- Noted the app & test targets in the project.  Verified that project modifications allow these to run smoothly (not quite sure that the app does besides function as a host for the tests).

### Impact to Consumers

- Since the podspec lists source files, current consumers will be able to continue consuming swift-wisdom without any visible impact.
- Updated `README` to include Carthage consumption instructions.

### Impact to Contributors

Because CocoaPods touches the workspace & project file (Pods-project and build-phases), to allow consumption via Carthage, these had to be removed to avoid build errors.  This means that **contributors must contribute via Carthage going forward instead of Cocoapods**.  Notably,

- when getting started, instead of cloning this repo and doing a `pod install`, one would do `carthage bootstrap`.
- when updating dependencies, instead of updating the `Podfile`, one would update the `Cartfile` instead.

### Impact to Workflows

These changes will likely break custom build/business workflows, or custom CI setups.  I don't have any visibility into these workflows, so any feedback is appreciated.


Steven-Wu-ConnectedLab:
CI failed as expected?

paulrolfe:
@Steven-Wu-ConnectedLab Hey! Nice to hear from you! This PR is not forgotten. We just need to get together and think about implications of using Carthage moving forward. We'll get back to you with some reviews/questions soon. Cheers.

PR #160, opened 2019-01-14
Remove one deprecated API

tzm41:
- Remove one deprecate API: `openURL`.
- Bump deployment target to iOS 10. From now on newer version of Intrepid pod will support minimum iOS 10.


SwiftWisdom/Core/UIKit/UIApplication+Extensions.swift
```diff
@@ -17,6 +17,6 @@ extension UIApplication {
         guard let url = URL(string: UIApplication.openSettingsURLString), canOpenURL(url) else {
             throw OpenSettingsError.cannotOpenURL
         }
-        openURL(url)
+        open(url)
```
bobgilmore, line 20: Let me investigate a bit and see if this is acceptable for extensions?

paulrolfe (commented):
Can this bump to remove support for 9.0 be captured in the readme? Similar to how we deal with swift versions?

jpmendel:
Yeah, I agree with @paulrolfe, we should probably start tracking the iOS version support in the readme. Otherwise this looks good.

tzm41:
@jpmendel @paulrolfe Added a tiny snip in README for iOS version.

jpmendel (approved):

bobgilmore (approved):

PR #161, opened 2019-02-19
Replace ip_round extension on Double for BinaryFloatingPoint

MaxHasADHD:
`ip_round` now works for Float


bobgilmore (changes requested):
Please add to the test, so that we see it explicitly exercised on both `Float`s and `Double`s

SwiftWisdom/Core/StandardLibrary/Numbers/BinaryFloatingPoint+Extensions.swift
```diff
@@ -0,0 +1,18 @@
...
+//
+//  Created by Stephen Wong on 5/1/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+
+extension BinaryFloatingPoint {
+    // https://codereview.stackexchange.com/questions/142748/swift-floatingpoint-rounded-to-places
+    public func ip_round(toDecimalPlaces places: Int) -> Self {
```
paulrolfe: Should it be `ip_rounded`? I think `ip_round` indicates it's a mutating function.

We could deprecate `ip_round` and forward it to an `ip_rounded` function.

paulrolfe (changes requested):

SwiftWisdom/Core/StandardLibrary/Numbers/BinaryFloatingPoint+Extensions.swift
```diff
@@ -0,0 +1,18 @@
...
+//  SwiftWisdom
+//
+//  Created by Stephen Wong on 5/1/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+
+extension BinaryFloatingPoint {
+    // https://codereview.stackexchange.com/questions/142748/swift-floatingpoint-rounded-to-places
```
paulrolfe: Good to cite your source. Can you also put in quick help comments?

SwiftWisdom/Core/StandardLibrary/Numbers/BinaryFloatingPoint+Extensions.swift
jpmendel: Or leave this one, make it `mutating` and add the new `ip_rounded`

jpmendel: Which I guess we would need to deprecate anyway since its a different type of function

SwiftWisdom.xcodeproj/project.pbxproj
```diff
@@ -938,21 +936,6 @@
 			shellScript = "diff \"${PODS_PODFILE_DIR_PATH}/Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [ $? != 0 ] ; then\n    # print error to STDERR\n    echo \"error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\" >&2\n    exit 1\nfi\n# This output is used by Xcode 'outputs' to avoid re-running this script phase.\necho \"SUCCESS\" > \"${SCRIPT_OUTPUT_FILE_0}\"\n";
 			showEnvVarsInLog = 0;
 		};
-		77882022D65A5916A7FFD999 /* [CP] Copy Pods Resources */ = {
```
bobgilmore, line 941: Why did this change? Did you re-run `pod install` If so, you could probably check in the `Podfile.lock`

SwiftWisdom/Core/StandardLibrary/Numbers/BinaryFloatingPoint+Extensions.swift
paulrolfe: Good point, might be nice to have the mutating version... would you mind adding that here too @MaxHasADHD? (I also don't mind turning that into a separate PR)

I don't know if swift has a way of deprecating and then repurposing a function name... I think the error that will be thrown from using a func with the wrong signature will be enough to indicate to the developer what to do. Other thoughts?

bobgilmore: Maybe mark the original version (with the now-bad signature) with some flavor of the `@available` attribute to show that it's no longer available?

bobgilmore: Based on https://docs.swift.org/swift-book/ReferenceManual/Attributes.html , maybe something like

```swift
@available(*, unavailable, renamed: "ip_rounded(toDecimalPlaces places: Int) -> Self")
public func ip_round(toDecimalPlaces places: Int) -> Self {
    fatalError("ip_round replaced by ip_rounded")
}
```
The string in the `@available` annotation isn't a valid Obj-C-like method signature. Clean it up as necessary.

SwiftWisdom.xcodeproj/project.pbxproj
MaxHasADHD: I did run `pod install` to generate the workspace

bobgilmore: Hmmm, since that made THIS change, didn't it also update `Podfile.lock`? You should rerun it and check in the lock file if it changed.

MaxHasADHD: Looks like the lock file did not change.

SwiftWisdom/Core/StandardLibrary/Numbers/BinaryFloatingPoint+Extensions.swift
```diff
@@ -0,0 +1,25 @@
...
+//  SwiftWisdom
+//
+//  Created by Stephen Wong on 5/1/16.
+//  Copyright © 2016 Intrepid. All rights reserved.
+//
+
+import Foundation
+
+extension BinaryFloatingPoint {
+    /// https://codereview.stackexchange.com/questions/142748/swift-floatingpoint-rounded-to-places
```
jpmendel, line 12: We might want to give this link the `- SeeAlso:` tag in the documentation along with adding a basic summary of the function.

SwiftWisdom/Core/StandardLibrary/Numbers/Double+Extensions.swift
```diff
@@ -2,13 +2,14 @@
...
+//  Created by Litteral, Maximilian on 2/22/19.
+//  Copyright © 2019 Intrepid. All rights reserved.
 //
 
 import Foundation
 
 extension Double {
+    @available(*, unavailable, renamed: "ip_rounded(toDecimalPlaces:)")
     public func ip_round(toDecimalPlaces decimalPlaces: Int) -> Double {
         let decimalShiftBase10 = pow(10.0, Double(decimalPlaces))
```
bobgilmore, line 14: Since this code now cannot be called, it should be removed. But since the method return a value, it needs to call `fatalError` or something similar.

bobgilmore (changes requested):

PR #172, opened 2020-01-13
[DPIOS-4] Audit of Swift-Wisdom

Nathaniel-Cox-Intrepid:
## Summary
Initial audit of Swift-Wisdom to remove out-of-date code. This was a fairly conservative pruning of low-hanging fruit.
Jira ticket: https://intrepid.jira.digital.accenture.com/browse/DPIOS-4

#### Removed:
- `IPResult` and tests

#### Marked as unavailable:
- `Array.ip_passes`: duplicate of `allSatisfy` in stdlib
- `Set.ip_passes`: duplicate of `allSatisfy` in stdlib
- `Set.ip_filter`: duplicate of `filter` in stdlib

#### Added `TODO` comments:
- `Array.ip_uniqueValues`: Inefficient implementation 
    - EDIT: updated implementation, now requires `Element` to be `Hashable`
- `String.ip_dropFirst`: propose renaming to `ip_removeFirst` to be more consistent with Apple API naming conventions 
    - EDIT: renamed. `ip_dropFirst()` has been marked unavailable.
- `UIColor.init(hex:)`: propose adding documentation or making failable init
    - EDIT: made into failable init

#### Updated implementation:
- `String.range(from:)`
- `UIView.ip_fromNib`

#### Didn't touch yet but would like to remove/deprecate:
- `Array.ip_generator` mostly duplicates the functionality of `makeIterator()` and is only needed if we need a type eraser that returns `AnyIterator` instead of `IndexingIterator<Element>`
-   `String.subscript(range: CountableRange<Int>)` Can be removed if Xcode 9 is no longer supported (I'm guessing yes?)
-   `String.ip_length` is this really necessary? Do people like `ip_length` better than using `count` directly?


ipbuildserver:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 169 tests, with 0 failures (0 unexpected) in 0.460 (0.521) seconds</td>
    </tr>
  </tbody>
</table>

</p>

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
```diff
@@ -53,6 +53,8 @@ extension String {
         }
     }
 
+    // TODO: Rename to ip_removeFirst()?
+    // In std library drop() returns a new String rather than mutating in place
     public mutating func ip_dropFirst() {
```
paulrolfe, line 57: Should we make this unavailable then? And point users to the swift functions `dropFirst` and `removeFirst`?

SwiftWisdom/Core/UIKit/UIColor+Hex.swift
```diff
@@ -9,6 +9,7 @@
 import UIKit
 
 public extension UIColor {
+    // TODO: document default behavior for malformed strings (or refactor to failable init)
     convenience init(ip_hex hex: String) {
```
paulrolfe: Let's just change this to an optional init. I like that better.

SwiftWisdom/Core/StandardLibrary/Array/Sequence+Utilities.swift
```diff
@@ -54,6 +54,7 @@ extension Sequence where Iterator.Element: Equatable {
     /// Provides only the unique values within a `Sequence`. ie: "1, 2, 2, 3" will return "1, 2, 3".
     ///
     /// - returns: Returns a sequence of the unique values
+    // TODO: Refactor to use Set for buffer?
```
paulrolfe: Not sure we want to do that. While this implementation is slow, it does maintain order, which I view as a benefit. It might be better just to document that feature of the function.

SwiftWisdom/Core/UIKit/UIView+Nib.swift
```diff
@@ -44,6 +37,7 @@ public extension UIView {
         let name = "\(self)".components(separatedBy: ".").first ?? ""
         return name
     }
+
```
paulrolfe: I see a trailing white space violation here

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
Nathaniel-Cox-Intrepid: That would be my thinking. `removeFirst` in the standard library crashes if called on an empty, so I could imagine there might still be a use case for this function along the lines of the safely indexing subscripts. But then it would more logically be called `ip_safelyRemoveFirst()` and return a `Character?` with `@discardable result`.

SwiftWisdom/Core/UIKit/UIColor+Hex.swift
Nathaniel-Cox-Intrepid: Ok, sounds good. This means making some adjustments to `ColorDescriptor` as well, since it currently uses this `init` for its `color` computed property.

SwiftWisdom/Core/StandardLibrary/Array/Sequence+Utilities.swift
Nathaniel-Cox-Intrepid: Excellent point. I've added to the documentation.

Nathaniel-Cox-Intrepid: Though it would be pretty easy to maintain order and O(n) time complexity:
```
var buffer = [Element]()
var seen = Set<Element>()
for element in self where !seen.contains(element) {
  buffer.append(element)
  seen.insert(element)
}
return buffer
```

SwiftWisdom/Core/StandardLibrary/String/String+Indexing.swift
paulrolfe: I like it

SwiftWisdom/Core/UIKit/UIColor+Hex.swift
paulrolfe: Does that over complicate it? I think it's a better API. Callers should be prepared to provide their own default color instead of having us give it to them.

SwiftWisdom/Core/StandardLibrary/Array/Sequence+Utilities.swift
paulrolfe: Love it.

SwiftWisdom/Core/UIKit/UIColor+Hex.swift
Nathaniel-Cox-Intrepid: I made some changes to `ColorDescriptor` that return `nil` instead of `fatalError` with bad input for its `init(_ value: String)`. It's not possible to make initializers from string literals failable, so I had to keep the fatal error there.

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
```diff
@@ -36,15 +36,15 @@ class String_IndexingTests: XCTestCase {
         XCTAssert("hello"[0..<5] == "hello")
     }
 
-    func testIpDropFirstWithNonEmptyString() {
+    func testIpSafelyRemoveFirstWithNonEmptyString() {
         var testString = "Hello"
-        testString.ip_dropFirst()
+        testString.ip_safelyRemoveFirst()
```
paulrolfe: Should you capture this and also test the return value? Since that's a thing that was added.

SwiftWisdom/Core/StandardLibrary/Array/Sequence+Utilities.swift
```diff
@@ -38,6 +38,29 @@ extension Sequence {
 }
 
 extension Sequence where Iterator.Element: Equatable {
+    /// Finds the count of a certain element.
+    ///
+    /// - parameter element: The element to be counted
+    ///
+    /// - returns: Number of instances of the element within the set
+    public func ip_count(ofElement element: Iterator.Element) -> Int {
+        return self.filter { $0 == element } .count
```
paulrolfe: Nit (and I know you didn't actually write this): Can you remove the space between `element }` and `.count`? It looks weird to me.

paulrolfe (commented):
Looks good, small suggestions. Also, do you know what's up with the Jenkins failures?

SwiftWisdomTests/StandardLibrary/String/String+IndexingTests.swift
Nathaniel-Cox-Intrepid: Good point. Thanks.

SwiftWisdom/Core/StandardLibrary/Array/Sequence+Utilities.swift
Nathaniel-Cox-Intrepid: Yes, it doesn't look weird. Thanks for catching that!

paulrolfe (approved):

PR #173, opened 2020-03-10
Fix bug in ip_fromNib()

Nathaniel-Cox-Intrepid:
This PR fixes the "fix" I made in the previous PR. My aim way to simplify the logic of the `ip_fromNib` function, which looked pretty ungainly. Here I take this one step further and completely do away with the optionality of this function. The optionality was a bit of a false-friend to begin with, as `Bundle.loadNibNamed(:owner:options:)` actually throws a runtime error if it can't find a nib instead of returning an empty array. I've tested this in a sample app, and I'm reasonable sure that it works with all, some or none of the optional parameters, at least for the testing I did.

### DIFF
--- a/SwiftWisdom/Core/UIKit/UIView+Nib.swift
+++ b/SwiftWisdom/Core/UIKit/UIView+Nib.swift
@@ -9,18 +9,15 @@
 import UIKit
 
 public extension UIView {
-    class func ip_fromNib<T: UIView>(_ nibNameOrNil: String? = nil, type: T.Type = T.self, in bundle: Bundle = Bundle.main) -> T {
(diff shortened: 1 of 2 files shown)


ipbuildserver:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">Executed 168 tests, with 0 failures (0 unexpected) in 0.429 (0.481) seconds</td>
    </tr>
  </tbody>
</table>

</p>

jpmendel (approved):

PR #175, opened 2020-08-11
RxSwift / RxCocoa update to 5.1.1 / Deployment Target to 10.0

John-Wolfe-Bose:
Got a warning from Apple that new app submissions will not be allowed to access UIWebView at all, and existing apps will not be able to submit new versions which access the UIWebView API starting in December 2020.

RxCocoa version 4.5 has a few extensions to UIWebView, and it seems like those extensions trigger the Apple check.

This pull request does several things:

* Update RxSwift / RxCocoa to version 5.1.1
* Add SwiftLint into the Podfile so it can be loaded locally
* Update SwiftLint build step for sample app to run first locally, then globally, then complain to user about missing SwiftLint
* Fixed SwiftLint orphaned_doc_comment warning
* Fix a BUNCH of RxSwift errors:
 ** 'Variable' replaced with 'BehaviorRelay'
 ** 'E' replaced with 'Element'
 ** Errors around DispatchTimeInterval initialization no longer assuming seconds (i.e., "5" changed to ".seconds(5)")
 ** 'next' deprecated
* Updated Intrepid Podspec to 0.16.0, deployment target to iOS 12.0
* Updated IntrepidTesting Podspec to 0.11.0, deployment target to iOS 12.0
* Ignore any .DS_Store files

### DIFF
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/IntrepidPursuits/swift-wisdom.git", :tag => "#{s.version}" }
   s.exclude_files = "tests/**/*"
   s.platform      = :ios
-  s.ios.deployment_target = "12.0"
+  s.ios.deployment_target = "10.0"
   s.pod_target_xcconfig = { 'SWIFT_VERSION' => '5.0' }
   s.default_subspec = "Core"
   s.swift_versions = ['4.2', '5.0']
--- a/Intrepid.podspec
+++ b/Intrepid.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "Intrepid"
-  s.version       = "0.15.1"
+  s.version       = "0.16.0"
   s.summary       = "Swift Bag"
   s.description   = <<-DESC
                     Collection of extensions and utility classes by and for the developers at Intrepid Pursuits.
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/IntrepidPursuits/swift-wisdom.git", :tag => "#{s.version}" }
   s.exclude_files = "tests/**/*"
   s.platform      = :ios
-  s.ios.deployment_target = "10.0"
+  s.ios.deployment_target = "12.0"
   s.pod_target_xcconfig = { 'SWIFT_VERSION' => '5.0' }
   s.default_subspec = "Core"
   s.swift_versions = ['4.2', '5.0']
@@ -37,7 +37,7 @@ Pod::Spec.new do |s|
 
   s.subspec "Rx" do |rx|
     rx.source_files = "SwiftWisdom/Rx/**/**/*.swift"
-    rx.dependency 'RxSwift', '~> 4.5'
-    rx.dependency 'RxCocoa', '~> 4.5'
+    rx.dependency 'RxSwift', '~> 5.1.1'
+    rx.dependency 'RxCocoa', '~> 5.1.1'
   end
 end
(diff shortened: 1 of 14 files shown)


Intrepid.podspec
```diff
@@ -25,7 +25,7 @@ Pod::Spec.new do |s|
   s.source        = { :git => "https://github.com/IntrepidPursuits/swift-wisdom.git", :tag => "#{s.version}" }
   s.exclude_files = "tests/**/*"
   s.platform      = :ios
-  s.ios.deployment_target = "10.0"
+  s.ios.deployment_target = "12.0"
```
andrewdolce: Have not yet done a full review pass, but just initially want to discuss why the deployment target is being increased. I know firsthand of some specific downstream apps that specifically require iOS 10 support (yes, I know that is generally overkill). So just want to think this through.

andrewdolce (changes requested):

SwiftWisdom/Rx/Rx+Extensions.swift
```diff
@@ -25,45 +25,61 @@ infix operator <-> : Binding
...
-public func <- <T: ObserverType, O>(observer: T, variable: Variable<O>) -> Disposable where T.E == O? {
-    return observer <- variable.asObservable()
+public func <- <T: ObserverType, O>(observer: T, variable: O) -> Disposable where T.Element == O? {
+    return observer <- BehaviorRelay(value: variable)
 }
 
-public func <- <T, O: ObservableType>(variable: Variable<T>, observable: O) -> Disposable where O.E == T {
-    return observable.bind(to: variable)
-}
+//public func <- <T, O: ObservableType>(variable: T, observable: O) -> Disposable where O.Element == T {
```
Nathaniel-Cox-Intrepid: Is there a reason to comment out this code rather than just deleting it?

SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
```diff
@@ -9,27 +9,27 @@ class OperatorTests: XCTestCase {
     func testVariableBinding() {
         let disposeBag = DisposeBag()
         let label = UILabel()
-        let variable = Variable<String?>(nil)
+        let variable = BehaviorRelay<String?>(value: nil)
```
Nathaniel-Cox-Intrepid: Tiny thing, but I might consider renaming the bindings in these tests to `relay` or `behaviorRelay` instead of `variable`.

SwiftWisdom/Rx/Rx+Extensions.swift
John-Wolfe-Bose: Just my paranoia - but the commented out code is deleted now

SwiftWisdomTests/Rx/Rx+ExtensionsTest.swift
John-Wolfe-Bose: Not so tiny - renamed and removed duplicated tests!

Intrepid.podspec
John-Wolfe-Bose: Rolled back deployment_targets to 10.0

Nathaniel-Cox-Intrepid (approved):
It looks good to me. Thanks for making these changes John!

John-Wolfe-Bose:
@Nathaniel-Cox-Intrepid / @andrewdolce - how will you guys accept this PR? Will it be in a new branch rx_5.1.1 ? Or something different than that?

John-Wolfe-Bose:
@andrewdolce - are you OK with these changes? 

@andrewdolce / @Nathaniel-Cox-Intrepid - how do you want to land these changes? Into a new branch? Into a new release? 

I am just circling around to make sure this PR does not get lost in the mists of time!

